PR #27, opened 2021-11-10
PLAT-484 Fix message when changing languages

daniel-klose:


oliver-richers (approved):
assuming you tested this

PR #29, opened 2021-11-11
PLAT-555 Add 'Releases and Versioning' section

alexander-schmidt:
- Referred to Semantic Versioning, and added examples.
- Unrelated: Swapped the latter two sections, added section numbers, and fixed some capitcalization.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ major  |  patch
      minor
 ```
 
-1. If there was an **API break** since the previous release, the **major version** in incremented: `1.2.3 -> 2.0.0` -- API breaks include:
+1. If there was an **API break** since the previous release, the **major version** is incremented: `1.2.3 -> 2.0.0` -- API breaks include:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,13 +53,33 @@ http://localhost:8000/portal/service?id=95cf1a1b-c12e-4594-9d20-783988fe32b9
...
 
-## Operations
+```
+     1.2.3
+    /  |  \
+major  |  patch
+     minor
+```
+
+1. If there was an **API break** since the previous release, the **major version** in incremented: `1.2.3 -> 2.0.0` -- API breaks include:
```
oliver-richers, line 67:
```suggestion
1. If there was an **API break** since the previous release, the **major version** is incremented: `1.2.3 -> 2.0.0` -- API breaks include:
```

oliver-richers (changes requested):

oliver-richers (approved):

PR #32, opened 2021-11-11
PLAT-521 Adapt WorkflowTaskManager:closeAllPreviousTasks

daniel-klose:


platform-workflow-module/src/main/java/com/softicar/platform/workflow/module/workflow/task/WorkflowTaskManager.java
```diff
@@ -43,6 +43,18 @@ public void setNextNodeAndGenerateTasks(AGWorkflowNode nextNode) {
 		}
 	}
 
+	public void closeAllOpenTasks() {
```
alexander-schmidt, line 46: since this method can now be called externally, it needs a transaction

alexander-schmidt (changes requested):

oliver-richers (changes requested):

oliver-richers (approved):

PR #43, opened 2021-11-22
PLAT-571 Do not enforce artifact signing

alexander-schmidt:
Allows for unsigned snapshot releases to a local Maven repository.

Test Plan:
- Edit `~/.gradle/gradle.properties`
  - Remove or comment `signing.keyId`
  - Set `ossrhUrl=/home/<user>/.softicar/maven-local`
- Run `./gradlew clean publish -Pversion=10.0.99-SNAPSHOT`
- Ensure that the build succeeds, and that artifacts are saved to the `ossrhUrl` directory.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -134,6 +134,7 @@ subprojects {
 def OSSRH_URL = findProperty('ossrhUrl')
 def OSSRH_USERNAME = findProperty('ossrhUsername')
 def OSSRH_PASSWORD = findProperty('ossrhPassword')
+def SIGNING_KEY_DEFINED = findProperty('signing.keyId') != null
 
 subprojects {
 	apply plugin: 'maven-publish'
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -190,15 +191,17 @@ subprojects {
...
-	publishMavenJavaPublicationToMavenRepository.doFirst {
-		if(OSSRH_URL == null) {
-			throw new GradleException("Please specify the variable 'ossrhUrl' in '~/.gradle/gradle.properties'.")
+	if(SIGNING_KEY_DEFINED) {
+		signing {
+				sign publishing.publications.mavenJava
 		}
-		if(version == "unspecified") {
-			throw new GradleException("Please specify the version to release with '-Pversion=X.Y.Z'.")
+		publishMavenJavaPublicationToMavenRepository.doFirst {
```
oliver-richers, line 198: why would we disable this, if no signing key is defined?

build.gradle
```diff
@@ -190,15 +191,17 @@ subprojects {
 			}
 		}
 	}
-	signing {
-		sign publishing.publications.mavenJava
-	}
-	publishMavenJavaPublicationToMavenRepository.doFirst {
-		if(OSSRH_URL == null) {
-			throw new GradleException("Please specify the variable 'ossrhUrl' in '~/.gradle/gradle.properties'.")
+	if(SIGNING_KEY_DEFINED) {
```
oliver-richers, line 194: hmm, dropping signing like this could lead to ppl forgetting it and then we publish without singing erroneously

oliver-richers (changes requested):

PR #50, opened 2021-11-29
PLAT-576 Add Notification when delegating Workflow Tasks

daniel-klose:


daniel-klose:
image

daniel-klose:
image

daniel-klose:
image

daniel-klose:
image

platform-workflow-module/src/main/java/com/softicar/platform/workflow/module/workflow/task/WorkflowTaskNotificationSubmitter.java
```diff
@@ -4,23 +4,32 @@
...
 	private final AGWorkflowTask task;
+	private final AGUser user;
 
 	public WorkflowTaskNotificationSubmitter(AGWorkflowTask task) {
 
 		this.task = task;
+		this.user = task.getUser();
+	}
+
+	public WorkflowTaskNotificationSubmitter(AGWorkflowTask task, AGUser user) {
```
alexander-schmidt, line 21: I think we should take a slightly different approach here, and replace this new constructor with a method:
```
public WorkflowTaskNotificationSubmitter setNotificationRecipient(AGUser notificationRecipient) {
    this.notificationRecipient = Objects.requireNonNull(notificationRecipient);
    return this;
}
```
The above constructor would still assign `this.notificationRecipient = task.getUser();`, ofc.

platform-workflow-module/src/main/java/com/softicar/platform/workflow/module/workflow/task/delegation/WorkflowTaskDelegationSaveHook.java
```diff
@@ -0,0 +1,19 @@
+package com.softicar.platform.workflow.module.workflow.task.delegation;
+
+import com.softicar.platform.emf.table.listener.IEmfSaveHook;
+import com.softicar.platform.workflow.module.workflow.task.WorkflowTaskNotificationSubmitter;
+import java.util.Collection;
+
+public class WorkflowTaskDelegationSaveHook implements IEmfSaveHook<AGWorkflowTaskDelegation> {
```
alexander-schmidt, line 7: I think this needs unit test coverage.

alexander-schmidt (changes requested):

platform-workflow-module/src/main/java/com/softicar/platform/workflow/module/workflow/task/delegation/WorkflowTaskDelegationSaveHook.java
daniel-klose: I dont think so.
The saveHooks themselves should already have unit tests, and all im doing here is using an already tested class.
As such, I dont really know what I should test here specifically

daniel-klose: But I added another test case in WorkflowTaskNotificationSubmitterTest

alexander-schmidt: I think that's exactly what should be tested.

alexander-schmidt (approved):

PR #51, opened 2021-11-29
PLAT-330 Fix Workflow Predicate Title

daniel-klose:


platform-workflow-module/src/main/java/com/softicar/platform/workflow/module/WorkflowI18n.java
```diff
@@ -295,6 +295,8 @@
 		.de("Arbeitsablauf-Benutzer");
 	I18n0 WORKFLOW_VERSION = new I18n0("Workflow Version")//
 		.de("Arbeitsablauf-Version");
+	I18n0 WORKFLOW_VERSION_DRAFT = new I18n0("Workflow Version Draft")//
+		.de("Entwurf der Arbeitsablauf-Version");
```
alexander-schmidt, line 299: I see no point in having a prefix for this predicate. Should simply be `DRAFT`.

alexander-schmidt (changes requested):

alexander-schmidt (approved):

PR #53, opened 2021-11-29
PLAT-100 Fix setPlaceholder for some IEmfInputs

daniel-klose:


daniel-klose:
The Problem described in PLAT-100 references the class `BigDecimalInputWithOptionalPlaceholder`

alexander-schmidt (changes requested):
I think we need to change our CSS now, and ditch the right-alignment of numerical inputs.

Rationale: Imagine a `DomLabelGrid` with a couple of stacked inputs, and some of those having lift-aligned placeholders while the rest would have right-aligned placeholders. This would look incredibly poor.

daniel-klose:
image
Now looks like this

alexander-schmidt (approved):

alexander-schmidt (approved):

PR #56, opened 2021-11-30
PLAT-565 Set systemUser as currentUser before executing programs

thees-k:
Test plan:

To reproduce the error add to `BufferedEmailSendProgram` in line 21:

```
System.out.println(CurrentUser.get().getLoginName());
```

When you start `BufferedEmailSendProgram` you will see that the login name of the logged-user is printed on console.

After the patch "system.user" will be printed out instead.

### DIFF
--- a/platform-core-module/src/main/java/com/softicar/platform/core/module/program/ProgramStarter.java
+++ b/platform-core-module/src/main/java/com/softicar/platform/core/module/program/ProgramStarter.java
@@ -4,6 +4,8 @@ import com.softicar.platform.common.core.exceptions.SofticarUserException;
 import com.softicar.platform.common.date.ISOCalendar;
 import com.softicar.platform.core.module.CoreI18n;
 import com.softicar.platform.core.module.log.LogDb;
+import com.softicar.platform.core.module.user.AGUser;
+import com.softicar.platform.core.module.user.CurrentUser;
 import com.softicar.platform.emf.source.code.reference.point.EmfSourceCodeReferencePoints;
 import java.util.UUID;
 
@@ -53,5 +55,6 @@ public class ProgramStarter {
 
 		ISOCalendar.setDefaultTimeZone();
 		LogDb.setProcessClass(program.getClass());
+		CurrentUser.set(AGUser.getSystemUser());
 	}
 }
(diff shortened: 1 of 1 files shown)


platform-core-module/src/main/java/com/softicar/platform/core/module/program/ProgramStarter.java
```diff
@@ -53,5 +55,6 @@ private void prepareExecution() {
 
 		ISOCalendar.setDefaultTimeZone();
 		LogDb.setProcessClass(program.getClass());
+		CurrentUser.set(AGUser.getSystemUser());
```
alexander-schmidt, line 58: I think we should use `CurrentUserScope` in the surrounding `try` block instead. E.g.
```
try (var scope = new CurrentUserScope(AGUser.getSystemUser())) {
	prepareExecution();
	program.executeProgram();
}
```
That would prevent problems when `start()` gets called directly from a UI thread.

alexander-schmidt (changes requested):

platform-core-module/src/main/java/com/softicar/platform/core/module/program/ProgramStarter.java
oliver-richers: I think the original solution was fine, considering the already existing code.

alexander-schmidt: I disagree. The proposed change eliminates a side effect which couly very easily be triggered by future code, while not really increasing complexity.

thees-k:
In my eyes both solutions show clearly why executor of a program is sytemUser.

But I don't care so I adjusted the code.

platform-core-module/src/main/java/com/softicar/platform/core/module/program/ProgramStarter.java
```diff
@@ -40,12 +42,9 @@ public static void main(String[] arguments) throws Throwable {
 
 	public void start() {
 
-		try {
+		try (var scope = new CurrentUserScope(AGUser.getSystemUser())) {
 			prepareExecution();
 			program.executeProgram();
-		} catch (Exception exception) {
```
oliver-richers, line 46: i don't think this `catch` should not be removed

oliver-richers (changes requested):

platform-core-module/src/main/java/com/softicar/platform/core/module/program/ProgramStarter.java
thees-k: I restored the catch block

oliver-richers (approved):
looks fine now

PR #67, opened 2021-12-07
PLAT-450 Using Integer Comparator in Data Table Priority Configuration

oliver-richers:


oliver-richers:
To test, open some management table with at least 10 columns, e.g. `AGCoreModuleInstance`. Open the table configuration popup and open the priority drop-down. Observe the order is now as expected for integers.
image

alexander-schmidt (approved):

PR #68, opened 2021-12-07
DFA-based CSV tokenizer (proof of concept)

oliver-richers:


oliver-richers:
was just a demonstration

PR #73, opened 2021-12-15
PLAT-608 Ensure deterministic row order during burst validation

alexander-schmidt:


platform-db-runtime/src/main/java/com/softicar/platform/db/runtime/table/listener/DbTableRowNotificationSet.java
```diff
@@ -22,7 +22,7 @@ public DbTableRowNotificationSet(IDbTable<R, ?> table) {
 	}
 
 	@Override
-	public Collection<DbTableRowNotification<R>> getAllNotifications() {
+	public List<DbTableRowNotification<R>> getAllNotifications() {
```
oliver-richers, line 25: why change this?

platform-db-runtime/src/main/java/com/softicar/platform/db/runtime/table/listener/IDbTableRowNotificationSet.java
```diff
@@ -14,36 +14,36 @@
  */
 public interface IDbTableRowNotificationSet<R> {
 
-	Collection<DbTableRowNotification<R>> getAllNotifications();
+	List<DbTableRowNotification<R>> getAllNotifications();
 
-	default Set<R> getAllTableRows() {
+	default List<R> getAllTableRows() {
```
oliver-richers, line 19: suggestion: return `Collection` here and everywhere below

platform-db-runtime/src/main/java/com/softicar/platform/db/runtime/table/listener/DbTableRowNotificationSet.java
```diff
@@ -1,19 +1,19 @@
...
 
 	private final IDbTable<R, ?> table;
-	private final Set<DbTableRowNotification<R>> notifications;
+	private final List<DbTableRowNotification<R>> notifications;
 
 	public DbTableRowNotificationSet(IDbTable<R, ?> table) {
 
 		this.table = table;
-		this.notifications = new HashSet<>();
+		this.notifications = new HashList<>();
```
oliver-richers, line 16: the set was preventing a duplication, if we change it to `HashList`, this is no longer the case

platform-db-runtime/src/main/java/com/softicar/platform/db/runtime/table/listener/IDbTableRowNotificationSet.java
```diff
@@ -14,36 +14,36 @@
...
 
 		return getAllNotifications()//
 			.stream()
 			.map(DbTableRowNotification::getTableRow)
-			.collect(Collectors.toCollection(HashSet::new));
+			.collect(Collectors.toCollection(HashList::new));
```
oliver-richers, line 24: here again, we remove the de-duplication when going to `List`

oliver-richers (changes requested):

platform-db-runtime/src/main/java/com/softicar/platform/db/runtime/table/listener/DbTableRowNotificationSet.java
alexander-schmidt: Yes it is. That's the whole point of `HashList`. See `HashList.add`:
```
	@Override
	public boolean add(T element) {

		if (set.add(element)) {
			list.add(element);
			return true;
		} else {
			return false;
		}
	}
```

platform-db-runtime/src/main/java/com/softicar/platform/db/runtime/table/listener/DbTableRowNotificationSet.java
alexander-schmidt: To imply that elements are contained in a specific order.

platform-db-runtime/src/main/java/com/softicar/platform/db/runtime/table/listener/DbTableRowNotificationSet.java
oliver-richers: you are right, I forgot about that :)

platform-db-runtime/src/main/java/com/softicar/platform/db/runtime/table/listener/DbTableRowNotificationSet.java
alexander-schmidt: (this change was reverted)

platform-db-runtime/src/main/java/com/softicar/platform/db/runtime/table/listener/IDbTableRowNotificationSet.java
alexander-schmidt: not correct; see above

oliver-richers (approved):
fine, but we should add some tests

PR #77, opened 2021-12-22
PLAT-612 Use flex layout for div elements

alexander-schmidt:
`div` elements are now styled with `display: flex` by default, as opposed to `display: grid`.

- All major browsers support `gap` in single-column `flex` layouts by now (cf. [caniuse.com](https://caniuse.com/?search=gap)), so we no longer need `grid` for that.
- This fixes abysmal performance in Firefox, which cannot efficiently handle deeply-nested grid layouts ([bug 1682686](https://bugzilla.mozilla.org/show_bug.cgi?id=1682686)).
- Also, it makes the Firefox DOM inspector usable, once again.
- By changing the default layout for `div`, the fix is not limited to the platform; it extends to complex UIs in consuming code as well.

Firefox performance **with this PR** _(watch this first!)_
firefox-performance-after

Firefox performance **before this PR** _(watch out for glitches and slowness... there's plenty of both)_
firefox-performance-before

Test Plan:
- Link `com.softicar.eas` against this code, and click through various pages and UIs.
- Note that there seem to be no visual regressions.


platform-dom/src/main/resources/com/softicar/platform/dom/elements/dom-day-input-style.css
```diff
@@ -1,4 +1,5 @@
 
 .DomDayInput {
 	display: inline-grid;
-}
\ No newline at end of file
+	grid-auto-columns: auto max-content;
```
alexander-schmidt, line 4: unrelated fix; the text input part of `DomDayInput` previously was not stretched horizontally, which looked poor in EMF input forms when displayed below regular text inputs

oliver-richers (approved):
assuming you tested this extensively, great improvement!!!

PR #80, opened 2022-01-03
PLAT-612 Use flex layout for button labels

alexander-schmidt:
Eliminates another widespread (and unnecessary) use of `display: grid`.

Test Plan:
Ensure that buttons look exactly like before this PR.


oliver-richers (approved):
ok

PR #81, opened 2022-01-04
PLAT-616 Do not decorate multi-file buttons in tables

alexander-schmidt:
Test Plan:
- Make sure that buttons for multi-file attributes are no longer decorated when displayed in a table.
- Test in Chromium and Firefox.

image

image


platform-dom/src/main/resources/com/softicar/platform/dom/elements/dom-button-style.css
```diff
@@ -14,7 +14,7 @@
 
 .DomButton {
 	/* position and layout */
-	justify-content: center;
+	align-items: center;
```
alexander-schmidt, line 17: semi-related fix; `justify-content` was wrong here before, and caused problems for this PR

daniel-klose (approved):

PR #82, opened 2022-01-05
PLAT-613 Close Popover when clicking actions

daniel-klose:


platform-emf/src/main/java/com/softicar/platform/emf/module/role/EmfModuleRoleViewButton.java
```diff
@@ -1,16 +1,29 @@
...
 		setIcon(EmfImages.USER_ROLE_ASSIGNMENT.getResource());
 		setLabel(EmfI18n.SHOW_USED_ROLES);
-		setClickCallback(() -> new EmfModuleRoleViewPopup(module).show());
+		setClickCallback(() -> {
+			callbackBeforeShow.ifPresent(INullaryVoidFunction::apply);
+			new EmfModuleRoleViewPopup(module).show();
+		});
+	}
+
+	public EmfModuleRoleViewButton setCallbackBeforeShow(INullaryVoidFunction callbackBeforeShow) {
```
alexander-schmidt, line 24: I think, we should instead use `DomPopupButton` here, and add `setCallbackBeforeShow` to it.

platform-core-module/src/main/java/com/softicar/platform/core/module/module/system/ModuleOverviewPageDiv.java
```diff
@@ -50,18 +50,29 @@ public ModuleOverviewPageDiv() {
...
+					.setLabel(AGModulePanicReceiver.TABLE.getPluralTitle())
+					.setTitle(EmfI18n.MANAGE_ARG1.toDisplay(AGModulePanicReceiver.TABLE.getPluralTitle())));
 			appendChild(
 				new DomButton()//
 					.setIcon(CoreImages.MODULES.getResource())
 					.setLabel(EmfI18n.SHOW_TABLES)
-					.setClickCallback(() -> new EmfTableOverviewPopup(module).show()));
-			appendChild(new EmfModuleRoleViewButton(module));
-			appendChild(new ModulePageOverviewButton(module));
+					.setClickCallback(() -> {
```
alexander-schmidt, line 63: We should probably extract this anonymous block to a private method.

alexander-schmidt (changes requested):

alexander-schmidt (approved):

PR #93, opened 2022-01-10
PLAT-637 Show Password in Popup after reset

daniel-klose:
image

### DIFF
--- a/platform-dom/src/main/resources/com/softicar/platform/dom/elements/dom-modal-dialog-style.css
+++ b/platform-dom/src/main/resources/com/softicar/platform/dom/elements/dom-modal-dialog-style.css
@@ -17,6 +17,7 @@
 	gap: var(--COLUMN_GAP);
 	min-width: 256px;
 	max-width: 512px;
+	white-space: pre-wrap;
 }
 
 .DomModalDialogPopupContent {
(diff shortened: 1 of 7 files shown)


alexander-schmidt (changes requested):
I think we should tweak the wording a bit, and add a paragraph in the message popup:
```
The password for user '%s' is now: %s

If an email server is configured, the user will receive this password via email.
```
...format args: login_name, password

German:
```
Das Passwort für Benutzer '%s' lautet jetzt: %s

Wenn ein E-Mail-Server konfiguriert ist, wird der Benutzer dieses Passwort via E-Mail erhalten.
```

alexander-schmidt (approved):

platform-dom/src/main/resources/com/softicar/platform/dom/elements/dom-modal-dialog-style.css
```diff
@@ -17,6 +17,7 @@
 	gap: var(--COLUMN_GAP);
 	min-width: 256px;
 	max-width: 512px;
+	white-space: pre-wrap;
```
alexander-schmidt, line 20: we may only do this for modal alerts, prompts and confirms - but not for modals in general.

alexander-schmidt (changes requested):

alexander-schmidt (approved):

PR #94, opened 2022-01-10
PLAT-638 Make DomButton Wrap its content properly

daniel-klose:
image

### DIFF
--- a/platform-dom/src/main/resources/com/softicar/platform/dom/elements/dom-button-style.css
+++ b/platform-dom/src/main/resources/com/softicar/platform/dom/elements/dom-button-style.css
@@ -21,7 +21,6 @@
 	flex-direction: row;
 	padding: 0;
 	position: relative;
-	white-space: nowrap;
 	justify-self: start;
 
 	/* look */


platform-dom/src/main/resources/com/softicar/platform/dom/elements/dom-button-style.css
```diff
@@ -21,7 +21,6 @@
 	flex-direction: row;
 	padding: 0;
 	position: relative;
-	white-space: nowrap;
```
alexander-schmidt, line 24: I don't think we should do this. Buttons with multi-line labels are _not_ a thing, in any UI I ever came across.

Instead, I think `white-space: nowrap;` should be retained, and `text-overflow: ellipsis;` should be added.

alexander-schmidt (changes requested):

platform-dom/src/main/resources/com/softicar/platform/dom/elements/dom-button-style.css
daniel-klose: Doesnt work unfortunately

alexander-schmidt:
superseded by https://github.com/softicar/platform/pull/95

PR #96, opened 2022-01-10
PLAT-618 Log the user who manually queues a program

daniel-klose:
image
image
image


alexander-schmidt (changes requested):
This whole approach might be overkill.
Couldn't we simply add a field `AGUser queuedBy` to `AGProgramExecution`, write it once upon record creation - and that's it?

platform-core-module/src/main/java/com/softicar/platform/core/module/program/execution/ProgramExecutionInserter.java
```diff
@@ -25,18 +27,24 @@ public AGProgramExecution insert() {
...
-
-		return program.getCurrentExecution();
+		try (var transaction = new DbTransaction()) {
+			program.reloadForUpdate();
+			program//
+				.setCurrentExecution(insertExecution(program.getQueuedBy()))
+				.setQueuedBy(null)
+				.setQueuedAt(null)
+				.save();
+			var currentExecution = program.getCurrentExecution();
```
alexander-schmidt, line 37: `currentExecution` could be inlined... won't make a difference, and looks less weird:
```
transaction.commit();
return program.getCurrentExecution();
```

platform-core-module/src/test/java/com/softicar/platform/core/module/program/AbstractProgramTest.java
```diff
@@ -0,0 +1,25 @@
...
+
+	public AbstractProgramTest() {
+
+		this.user = insertUser();
+	}
+
+	private AGUser insertUser() {
+
+		return new AGUser()//
+			.setActive(false)
```
alexander-schmidt, line 18: Why would we insert an _inactive_ user? This could (or even should) cause all sorts of problems.

platform-core-module/src/test/java/com/softicar/platform/core/module/program/execution/scheduled/ScheduledProgramEnqueuerTest.java
```diff
@@ -4,12 +4,12 @@
...
 import com.softicar.platform.common.date.Time;
 import com.softicar.platform.core.module.program.AGProgram;
+import com.softicar.platform.core.module.program.AbstractProgramTest;
 import com.softicar.platform.core.module.program.execution.AGProgramExecution;
-import com.softicar.platform.db.runtime.test.AbstractDbTest;
 import java.util.UUID;
 import org.junit.Test;
 
-public class ScheduledProgramEnqueuerTest extends AbstractDbTest {
+public class ScheduledProgramEnqueuerTest extends AbstractProgramTest {
```
alexander-schmidt, line 12: I think we need to modify `assertQueuedAt` below, as follows:
- rename it to `assertQueued`
- also let it check the `QUEUED_BY` field

Like so:
```
private void assertQueued(DayTime expectedQueuedAt) {

	assertEquals(expectedQueuedAt, program.getQueuedAt());
	assertEquals(expectedQueuedAt != null? /*system user*/ : null, program.getQueuedBy());
}
```

alexander-schmidt (changes requested):

alexander-schmidt (approved):

PR #101, opened 2022-01-17
PLAT-643 Move columns of AGProgram to new table AGProgramState

thees-k:
- Add AbstractTransientDayTimeField
- Tmp commit
- Fix the most compile errors
- Add AGProgramState and adjust code accordingly
- Remove dead code, simplify code


platform-core-module/src/main/java/com/softicar/platform/core/module/program/AGProgram.java
```diff
@@ -24,14 +30,54 @@ public AGProgram setProgramUuid(UUID programUuid) {
 		return setProgramUuid(AGUuid.getOrCreate(programUuid));
 	}
 
+	public DayTime getQueuedAt() {
+
+		return getOrCreateProgramState().getQueuedAt();
+	}
+
+	public void saveQueuedAt(DayTime value) {
```
alexander-schmidt, line 38: This method is _only_ called from unit tests. It should be removed (test-only parts in production code should commonly be avoided). If necessary, equivalent methods can be added to the respective test classes.

platform-core-module/src/main/java/com/softicar/platform/core/module/program/AGProgram.java
```diff
@@ -24,14 +30,54 @@ public AGProgram setProgramUuid(UUID programUuid) {
...
+
+		getOrCreateProgramState().setQueuedAt(value).save();
+	}
+
+	public AGUser getQueuedBy() {
+
+		return getOrCreateProgramState().getQueuedBy();
+	}
+
+	public void saveQueuedBy(AGUser value) {
```
alexander-schmidt, line 48: This method is _only_ called from unit tests. It should be removed (test-only parts in production code should commonly be avoided). If necessary, equivalent methods can be added to the respective test classes.

platform-core-module/src/main/java/com/softicar/platform/core/module/program/AGProgram.java
```diff
@@ -24,14 +30,54 @@ public AGProgram setProgramUuid(UUID programUuid) {
...
+
+		getOrCreateProgramState().setAbortRequested(value).save();
+	}
+
+	public AGProgramExecution getCurrentExecution() {
+
+		return getOrCreateProgramState().getCurrentExecution();
+	}
+
+	public void saveCurrentExecution(AGProgramExecution value) {
```
alexander-schmidt, line 68: This method is _only_ called from unit tests. It should be removed (test-only parts in production code should commonly be avoided). If necessary, equivalent methods can be added to the respective test classes.

platform-core-module/src/main/java/com/softicar/platform/core/module/program/AGProgram.java
```diff
@@ -24,14 +30,54 @@ public AGProgram setProgramUuid(UUID programUuid) {
...
+
+		getOrCreateProgramState().setQueuedBy(value).save();
+	}
+
+	public Boolean isAbortRequested() {
+
+		return getOrCreateProgramState().isAbortRequested();
+	}
+
+	public void saveAbortRequested(Boolean value) {
```
alexander-schmidt, line 58: This method is only called only once, from `ProgramAbortRequester` - and it's unlikely that further references will be added. I think we should inline/remove it.

platform-core-module/src/main/java/com/softicar/platform/core/module/program/AGProgram.java
```diff
@@ -106,20 +162,20 @@ public static AGProgram loadOrInsert(AGUuid programUuid) {
 
 	private static AGProgram executeLoadOrInsert(AGUuid programUuid) {
 
-		return AGProgram.TABLE//
-			.createSelect(SqlSelectLock.FOR_UPDATE)
```
alexander-schmidt, line 110: This eager `for-update` conditional lock was actually important: If we simply use `AGProgram.loadByProgramUuid` instead, multiple threads could try to concurrently insert `AGProgram` records with the same UUIDs.

This needs to be fixed/reverted.

platform-core-module/src/main/java/com/softicar/platform/core/module/program/AGProgramTable.java
```diff
@@ -33,6 +37,22 @@ public void customizeAttributeProperties(IEmfAttributeList<AGProgram> attributes
 			.setTitle(CoreI18n.PROGRAM)
 			.setImmutable(true)
 			.setPredicateMandatory(EmfPredicates.always());
+
+		attributes//
+			.addTransientAttribute(AGProgram.QUEUED_AT)
+			.setDisplayFactory(EmfDayTimeDisplay::new);
```
alexander-schmidt, line 43: Do we really need to explicitly define display factories here? Because this _should_ work out-of-the-box.
I am not certain though.

(same for other transient attributes below)

platform-core-module/src/main/java/com/softicar/platform/core/module/program/ProgramEnqueuer.java
```diff
@@ -34,7 +34,8 @@ private void enqueueExecution(AGUuid programUuid) {
 		AGProgram program = AGProgram.loadOrInsert(programUuid);
 		if (program.reloadForUpdate()) {
```
alexander-schmidt, line 35: Locking should be performed on the `AGProgramState` record instead, e.g.:
```
var state = AGProgram.loadOrInsert(programUuid).getState();
if (state.reloadForUpdate()) {
	state//
		.setQueuedAt(...)
		.setQueuedBy(...)
		.save();
}
```

platform-core-module/src/main/java/com/softicar/platform/core/module/program/ProgramEnqueuer.java
```diff
@@ -34,7 +34,8 @@ private void enqueueExecution(AGUuid programUuid) {
...
 			program//
-				.setQueuedAt(DayTime.now())
+				.getOrCreateProgramState()
+				.setQueuedAt(DayTime.now().truncateSeconds())
```
alexander-schmidt, line 38: We did not truncate seconds before - but we do now. Why so?

platform-core-module/src/main/java/com/softicar/platform/core/module/program/execution/ProgramExecutionInserter.java
```diff
@@ -27,16 +26,14 @@ public AGProgramExecution insert() {
...
-			program.reloadForUpdate();
-			program//
-				.setCurrentExecution(insertExecution(program.getQueuedBy()))
-				.setQueuedBy(null)
-				.setQueuedAt(null)
-				.save();
-			transaction.commit();
-			return program.getCurrentExecution();
-		}
+		program.reloadProgramStateForUpdate();
```
alexander-schmidt, line 29: I think this should be rewritten as:
```
var state = program.getState();
state.reloadForUpdate();
state//
    .setCurrentExecution(...)
    .set...
    .save();
```
...so that it's more obvious that this is all about the `state` record.

This, ofc, assumes that `reloadProgramStateForUpdate` was eliminated/inlined.

platform-db-runtime/src/main/java/com/softicar/platform/db/runtime/transients/AbstractTransientDayTimeField.java
```diff
@@ -0,0 +1,22 @@
+package com.softicar.platform.db.runtime.transients;
+
+import com.softicar.platform.common.date.DayTime;
+import com.softicar.platform.db.runtime.table.row.IDbTableRow;
+
+/**
+ * Defines a transient field with values of type {@link DayTime}.
+ * <p>
+ * Please note that null-values are not supported. If the values are optional,
```
alexander-schmidt, line 9: The Javadoc states that "null-values are not supported" - but yet we use this base class to implement `[Program]QueuedAtField`, which in turn refers to the nullable field `AGProgramState.QUEUED_AT`.

Something's odd or even wrong here.

platform-db-runtime/src/main/java/com/softicar/platform/db/runtime/transients/TransientFieldValueTypes.java
```diff
@@ -28,6 +29,11 @@
 		return getComparable(BigDecimal.class);
 	}
 
+	public static final ITransientFieldValueType<DayTime> getDayTime() {
```
alexander-schmidt, line 32: Wow, this should not have been missing here.

We should add methods (and base classes) for other missing types as well (e.g. `Day`)... but probably not in this PR.

alexander-schmidt (changes requested):

platform-core-module/src/main/java/com/softicar/platform/core/module/program/AGProgram.java
thees-k: I changed that.

platform-core-module/src/main/java/com/softicar/platform/core/module/program/AGProgram.java
thees-k: I changed that.

platform-core-module/src/main/java/com/softicar/platform/core/module/program/AGProgram.java
thees-k: I changed that.

platform-core-module/src/main/java/com/softicar/platform/core/module/program/AGProgram.java
thees-k: I changed that.

platform-core-module/src/main/java/com/softicar/platform/core/module/program/AGProgram.java
```diff
@@ -1,18 +1,24 @@
...
 import com.softicar.platform.core.module.uuid.AGUuid;
 import com.softicar.platform.db.core.transaction.DbTransactions;
-import com.softicar.platform.db.sql.statement.SqlSelectLock;
 import com.softicar.platform.emf.object.IEmfObject;
 import java.util.Objects;
 import java.util.UUID;
 
 public class AGProgram extends AGProgramGenerated implements IEmfObject<AGProgram> {
 
+	public static final AbortRequestedField ABORT_REQUESTED = new AbortRequestedField();
```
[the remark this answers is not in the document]
thees-k: I changed that.

platform-core-module/src/main/java/com/softicar/platform/core/module/program/AGProgramState.java
```diff
@@ -0,0 +1,128 @@
+package com.softicar.platform.core.module.program;
```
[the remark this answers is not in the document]
thees-k: I changed that.

platform-core-module/src/main/java/com/softicar/platform/core/module/program/AGProgram.java
```diff
@@ -92,11 +138,21 @@ public void resetAll() {
 	 */
 	public void resetCurrentExecution() {
 
-		getThis()//
+		getOrCreateProgramState()//
 			.setCurrentExecution(null)
 			.save();
 	}
 
+	public AGProgramState getOrCreateProgramState() {
```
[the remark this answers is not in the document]
thees-k: ok, it's changed!

platform-core-module/src/main/java/com/softicar/platform/core/module/program/AGProgram.java
```diff
@@ -92,11 +138,21 @@ public void resetAll() {
...
+
+		return AGProgramState.TABLE.getOrCreate(getThis());
+	}
+
+	public boolean reloadProgramStateForUpdate() {
```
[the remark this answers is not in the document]
thees-k: I renamed it to `reloadStateForUpdate()`.
I prefer delegation methods because of the OOP principle.

platform-core-module/src/main/java/com/softicar/platform/core/module/program/AGProgram.java
thees-k: Ok, thanks for info! I reverted it.

platform-core-module/src/main/java/com/softicar/platform/core/module/program/AGProgramTable.java
thees-k: Yes, it it necessary to define display factories.

Otherwise whole records are displayed:
Screenshot_2022-01-18_14-49-04
And note that column "Abort Requested" shows "false" instead of "No".

platform-core-module/src/main/java/com/softicar/platform/core/module/program/ProgramEnqueuer.java
thees-k: I changed that.

platform-core-module/src/main/java/com/softicar/platform/core/module/program/ProgramEnqueuer.java
thees-k: Until commit 92c9622 the seconds were truncated.
With cron syntax the user can start programs only at the full minute. I think this is why the originator of the code decided to not display seconds to the user.

platform-core-module/src/main/java/com/softicar/platform/core/module/program/execution/ProgramExecutionInserter.java
thees-k: I would like to keep method `reloadStateForUpdate()` because of delegation principle

platform-db-runtime/src/main/java/com/softicar/platform/db/runtime/transients/AbstractTransientDayTimeField.java
thees-k: Sorry, it was a copy-and-paste error! I fixed that.

platform-db-runtime/src/main/java/com/softicar/platform/db/runtime/transients/TransientFieldValueTypes.java
thees-k: I agree.

platform-core-module/src/main/java/com/softicar/platform/core/module/program/AGProgramTable.java
alexander-schmidt: Oh... so, our framework is less awesome than expected. Never mind.

platform-core-module/src/main/java/com/softicar/platform/core/module/program/execution/ProgramExecutionInserter.java
alexander-schmidt: I'd hold the KISS principle against this - but anyway... it's probably just a matter of taste.

platform-core-module/src/main/java/com/softicar/platform/core/module/program/ProgramEnqueuer.java
alexander-schmidt: Regardless of the history, I think that we should not truncate seconds here.
Otherwise, I am concerned that we might give ourselves a hard time, in concealing/faking the point in time at which a program was actually enqueued. Particularly, by discarding seconds, we're unable to tell whether there was a significant delay between the _intended_ (as defined by the schedule) and the _actual_ enqueuing of a program.
This kind of information might be relevant, not only to us as developers but also to an admin of a live instance.

thees-k: Ok, thanks for explanation!
I changed that.

alexander-schmidt (approved):

PR #103, opened 2022-01-18
PLAT-645 Fix Properties not Trimming their Values

daniel-klose:
image
image


platform-common-core/src/main/java/com/softicar/platform/common/core/properties/AbstractProperty.java
```diff
@@ -59,7 +59,7 @@ public T getValue() {
 
 		String valueString = getValueString(propertyName);
 		if (valueString != null) {
-			return parser.apply(valueString);
+			return parser.apply(valueString.trim());
```
alexander-schmidt, line 62: We need a unit test for this. (should have existed all along)

alexander-schmidt (changes requested):

alexander-schmidt (approved):

PR #104, opened 2022-01-20
PLAT-376 Refactor and encapsulate Smb related code

daniel-klose:
We have the following test setup:
```
testShare
|---testfolder
	|--- test1
	|--- test2
	|--- test3
	|--- testFolder2
		|--- test2_1
		|--- test2_2
		|--- test2_3
```
`listFiles` and `listFilesRecursively` on the `testFolder` show us this output:
image
Also tested renaming and moving files.


alexander-schmidt (changes requested):

platform-core-module/src/main/java/com/softicar/platform/core/module/file/smb/SmbCredentials.java
```diff
@@ -0,0 +1,33 @@
+package com.softicar.platform.core.module.file.smb;
+
+public class SmbCredentials implements ISmbCredentials {
```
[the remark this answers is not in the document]
daniel-klose: I have seriously no idea how to describe this class since it is too trivial

platform-core-module/src/main/java/com/softicar/platform/core/module/file/smb/jcifs/JcifsSmbDirectory.java
```diff
@@ -37,6 +41,28 @@ public JcifsSmbDirectory(String context, String name, NtlmPasswordAuthentication
 		}
 	}
 
+	@Override
+	public Collection<String> listFilesRecursively(String root) {
```
alexander-schmidt, line 45: Something went wrong here - `listFilesRecursively` and `listFiles` should not need to be so similar, and complicated.

This might be sufficient to walk the directory tree, in depth-first order. Did not test it though.
```
@Override
public Collection<String> listFilesRecursively(String subDirectory) {

	List<String> fileNames = new ArrayList<>();
	for (ISmbFile file: getSubDirectory(subDirectory).listFiles()) {
		if (file.isDirectory()) {
			fileNames.addAll(listFilesRecursively(subDirectory + "/" + file.getName()));
		} else {
			fileNames.add(file.getCanonicalPath());
		}
	}
	return fileNames;
}
```

platform-core-module/src/main/java/com/softicar/platform/core/module/file/smb/ISmbDirectory.java
```diff
@@ -16,6 +17,8 @@
 
 	List<ISmbFile> listFiles();
 
+	Collection<String> listFilesRecursively(String root);
```
alexander-schmidt, line 20: I think the parameter `root` should be renamed to `subDirectory`.

platform-core-module/src/main/java/com/softicar/platform/core/module/file/stored/content/store/StoredFileSmbContentStore.java
```diff
@@ -160,50 +113,26 @@ public long getFileSize(String filename) {
...
 
-		return new NtlmPasswordAuthentication(//
-			fileServer.getDomain(),
-			fileServer.getUsername(),
-			fileServer.getPassword());
+		AGStoredFileServer fileServer = getServerOrThrow();
+		return CurrentSmbClient
+			.get()
+			.createFile(
+				getServerOrThrow().getUrl() + "/" + name,
```
alexander-schmidt, line 134:
```suggestion
				Trim.trimRight(getServerOrThrow().getUrl(), '/') + "/" + name,
```
We should probably remove a tailing `/` from the return value of `getUrl`, it any, to make sure that we never end up with `//` in the resulting URL.

alexander-schmidt (changes requested):

platform-core-module/src/main/java/com/softicar/platform/core/module/file/smb/jcifs/JcifsSmbDirectory.java
daniel-klose: I dont think this will work, maybe im just confused because its recursion, but with the way you suggest, I have a feeling it wont work when I look at `StoredFileSmbContentStore` 
Hard to explain via text, but I reverted some other changes from me so that the implementation is now nearly the same as it was in `StoredFileSmbContentStore` which should reduce the chance that something broke

daniel-klose: Doesnt work, fixed and tested my version

platform-core-module/src/main/java/com/softicar/platform/core/module/file/smb/ISmbFile.java
```diff
@@ -46,6 +46,8 @@
 
 	ISmbFile moveTo(ISmbDirectory parent);
 
+	void copyTo(String path);
```
alexander-schmidt, line 49: `path` should be `url` - because a full URL needs to be provided here

platform-core-module/src/main/java/com/softicar/platform/core/module/file/smb/jcifs/JcifsSmbFile.java
```diff
@@ -151,6 +151,16 @@ public ISmbFile moveTo(ISmbDirectory parent) {
 		return moveAndRenameTo(parent, file.getName());
 	}
 
+	@Override
+	public void copyTo(String path) {
```
alexander-schmidt, line 155: `path` should be `url` - because a full URL needs to be provided here

alexander-schmidt (changes requested):

alexander-schmidt (approved):

PR #106, opened 2022-01-25
PLAT-647 Create EmfTokenMatrixParser

alexander-schmidt:
- Created `EmfTokenMatrixParser` to derive `IEmfTableRow` instances from a `List` of `List`s of `String` based tokens, and a specified target table.
- Created token-to-object converters for all 14 relevant DB table field types.
  - Those are: `BigDecimal`, `Boolean`, `ByteArray`, `Day`, `DayTime`, `Double`, `Enum`, `Float`, `Foreign`, `ID`, `Integer`, `Long`, `String`, `Time`
  - Available converters are registered in `EmfTokenConverters`.
- Basic validation of `String` based tokens is performed.
  - Mode of operation is "fail-fast".
  - Encountered error messages are i18n'ed.
  - Error messages are crafted to be as helpful as possible, and to quickly point to the offending token.
  - Error message wording aims towards a system administrator, or even trained-end-user, audience.

Test Plan:
- check API of `EmfTokenMatrixParser`
- see `EmfTokenRowParserTest` and `EmfTokenMatrixParserExceptionBuilderTest`


platform-common-testing/src/main/java/com/softicar/platform/common/testing/Asserts.java
```diff
@@ -71,11 +71,28 @@ public static void assertExceptionMessage(IDisplayString expectedMessage, INulla
 		try {
 			thrower.apply();
 			fail("An expected exception failed to occur.");
+		} catch (NullPointerException exception) {
```
alexander-schmidt, line 74: An NPE has `exception.getMessage() == null`. This could result in confisuing behavior of this assertion method. Hence the new distinct handling of NPE.

platform-emf/src/main/java/com/softicar/platform/emf/token/parser/converter/EmfTokenToByteArrayConverter.java
```diff
@@ -0,0 +1,22 @@
+package com.softicar.platform.emf.token.parser.converter;
+
+import com.softicar.platform.common.core.i18n.IDisplayString;
+import com.softicar.platform.db.runtime.field.DbByteArrayField;
+import com.softicar.platform.emf.EmfI18n;
+import java.util.Base64;
+
+class EmfTokenToByteArrayConverter extends AbstractEmfTokenConverter<byte[], DbByteArrayField<?>> {
```
alexander-schmidt, line 8: Note to myself: This needs to consider maxlength and lengthbits (much like the String converter).

daniel-klose (dismissed):

daniel-klose (approved):

PR #107, opened 2022-01-26
PLAT-623 Upgrade h2 to 2.1.210

alexander-schmidt:
All `1.x` versions have critical vulnerabilities that do not seem to get fixed anymore in the 1.x release series.
The current release, `2.1.210`, is the only one without known vulnerabilities.
This PR upgrades H2 to `2.1.210`.

Several H2 `2.x` changes were adopted in this PR:
- `value` is now a reserved keyword.
  - In `1.x`, it was not.
  - To compensate for that, various columns in H2-based unit tests were renamed.
- H2 no longer tolerates `VARCHAR(0)` in column definitions.
  - In `1.x`, `VARCHAR(0)` is interpreted as `VARCHAR(2147483647)`.
  - In `2.x`, `VARCHAR` is interpreted as `VARCHAR(2147483647)`. When `VARCHAR(0)` is specified, an exception occurs.
  - To compensate for that, yet another H2-specific quirk was added to `DbColumnStructureSqlGenerator`.
  - The same goes for `VARBINARY`.
- H2 now returns inserted PKs after a multi-insert, even if a PK was explicitly specified, and even if the PK is non-auto-increment.
  - That means, for auto-increment PKs, a longstanding H2 quirk was finally fixed upstream. Expectations in `DbTableInsertTest` were changed accordingly.
  - For non-auto-increment PKs (e.g. for composite PKs), a check was added to `DbConnection`: We no longer try to extract generated IDs from non-auto-increment columns.
- When casting to `CHAR`, H2 now assumes `CHAR(1)`.
  - To compensate for that, `DbResultSetCastingTest` now uses `VARCHAR` in its String-casting test.
- In `INFORMATION_SCHEMA.TABLES`, regular tables now have `TABLE_TYPE = 'BASE TABLE'`, instead of `TABLE_TYPE = 'TABLE'`.
  - This now matches the behavior of MariaDB.
  - Expectations in `DbTableNameFetcherTest` and `DbMetadataRequesterTest` were changed accordingly.
- `INFORMATION_SCHEMA` now defines various views.
  - That is, 20 actual views inside `INFORMATION_SCHEMA`. This is entirely new, and not to be confused with regular "payload table" view definitions in `INFORMATION_SCHEMA.VIEWS`.
  - To compensate for that, `DbViewStructuresLoader` now ignores views inside `INFORMATION_SCHEMA`, by default.
- In `INFORMATION_SCHEMA.VIEWS`, `VIEW_DEFINITION` no longer starts with `CREATE FORCE VIEW [...] AS [...]`.
  - This now matches the behavior of MariaDB.
  - Expectations in `DbViewStructuresLoaderTest` were changed accordingly.
- In `CREATE TABLE` statements, UK constraints now have to be defined via `UNIQUE`. Using `UNIQUE KEY` causes an exception.
  - To compensate for that, the H2 MySQL-Compatibility-Mode was enabled in `DbH2MemoryConnector`.
- H2 now seems to use a different default timezone than before.
  - `AbstractTest` now sets a default timezone, to ensure that the unit test itself and the JVM in the spawned H2 child process use the same timezone.

Test Plan:
- Ensure that this generates no output: `./gradlew allDependencies | grep FAIL`
- Ensure that all tests are green: `./gradlew clean check`
  - Note that a DB based unit tests use h2.


platform-db-core/src/main/java/com/softicar/platform/db/core/test/AbstractDbCoreTest.java
```diff
@@ -14,6 +14,8 @@
 
 	public AbstractDbCoreTest() {
 
+//		Log.finfo(System.getProperty("user.timezone"));
+
```
daniel-klose, line 18: this seems like a debug, remove if so

daniel-klose (changes requested):

platform-db-core/src/main/java/com/softicar/platform/db/core/test/AbstractDbCoreTest.java
alexander-schmidt: indeed

daniel-klose (approved):

PR #125, opened 2022-02-01
PLAT-651 Rework UserActivityPage

daniel-klose:
Removed UserActivityPage and instead added a transient field to AGUser
Unrelated: Cleaned up orphaned `CoreI18` entries
image


platform-core-module/src/main/java/com/softicar/platform/core/module/user/login/overview/UserLoginOverviewDiv.java
```diff
@@ -0,0 +1,64 @@
...
+			new DomActionBar(
+				new DomButton()//
+					.setIcon(EmfImages.REFRESH.getResource())
+					.setLabel(CoreI18n.REFRESH)
+					.setClickCallback(this::refresh)));
+		appendChild(container);
+		refresh();
+	}
+
+	private void refresh() {
```
alexander-schmidt, line 32: I think we should instead
- call `new UserLoginOverviewTableDivBuilder().build()` in the CTOR, and assign its return value (an `IEmfDataTableDiv`) to a field `dataTableDiv`
- change this `refresh` method to simply call ``dataTableDiv.refresh()`

This way, filters and other table settings won't get lost when somebody clicks the "Refresh" button.

alexander-schmidt (changes requested):
I wonder if all of this is worth the hassle. Do we even need a separate page for this? How about simply adding two transient fields to `AGUserTable`:
1. last login date
2. number of days since last login

daniel-klose:
Well, either we move all Login-Related Code to `AGUserTable` or dont move it at all.
Im completely against moving only parts of the Login-related code, either all nor nothing.

alexander-schmidt (changes requested):

alexander-schmidt (approved):

PR #134, opened 2022-02-08
PLAT-685 DbConnection: Fix a wrong assumption about ResultSet metadata

alexander-schmidt:
Contracy to the H2 driver, the MariaDB driver seems to not flag an RS column as auto-increment if, during an insert, a value was explicitly defined for that column.

Test Plan:
- Create a local snapshot release of the platform.
- Use that release in an EAS checkout.
- Create a war file from that EAS checkout.
- Deploy the war file on the Tomcat server of a SoftiCAR machine with persistent MariaDB.
- Open the login URL, and use your credentials to log in.
- Note that the login works (results in an error message w/o this PR).

### DIFF
--- a/platform-db-core/src/main/java/com/softicar/platform/db/core/connection/DbConnection.java
+++ b/platform-db-core/src/main/java/com/softicar/platform/db/core/connection/DbConnection.java
@@ -21,6 +21,7 @@ import java.sql.DatabaseMetaData;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.sql.Types;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


platform-db-core/src/main/java/com/softicar/platform/db/core/connection/DbConnection.java
```diff
@@ -248,10 +249,19 @@ public int executeInsert(DbStatement statement) {
...
 	}
 
-	private boolean isFirstColumnAutoIncrement(ResultSet resultSet) throws SQLException {
+	private boolean hasNumericFirstColumn(ResultSet resultSet) throws SQLException {
 
 		var metaData = resultSet.getMetaData();
-		return metaData.getColumnCount() > 0 && metaData.isAutoIncrement(1);
+		if (metaData.getColumnCount() > 0) {
+			switch (metaData.getColumnType(1)) {
+			case Types.BIGINT:
```
oliver-richers, line 257: hmm, in line 238, we call `getInt` which will truncate the BIGINT...maybe we should add some TODO or FIXME in 238, referring to our missing BIGINT-auto-increment support?

PR #142, opened 2022-02-15
PLAT-692 Fix Chaotic Row Order of Table of Program Page

thees-k:
#### Summary
- PLAT-692 Add `ProgramField` to `AGProgram` to be able to sort table by name
- Put program column behind ID column
- Reduce visibility of *Field classes (unrelated to PLAT-692)

#### Test Plan
See Program Page, it shows ID column, too, now, and therefore looks like other pages.
The bug was fixed: You can sort by Program column, now.
Filtering by program is also possible.
Screenshot_2022-02-15_09-33-16


oliver-richers (approved):
not the best solution, but does the job

PR #144, opened 2022-02-15
PLAT-581 Minor improvement to current ImportEngine

daniel-klose:
Dont show import button in tables where it is currently not possible to import entites for (ForeignRowAttributes)
Ignore concealed columns in import dialog
image


platform-emf/src/main/java/com/softicar/platform/emf/management/importing/engine/EmfImportEngine.java
```diff
@@ -39,6 +39,7 @@ public EmfImportEngine(IEmfTable<R, P, S> table) {
 		ignoreActiveField();
 		ignoreGeneratedFields();
 		ignoreTransactionFields();
+		ignoreConcealedFields();
```
oliver-richers, line 42: i am not sure if we can ignore concealed fields here, can we talk about some examples?

platform-emf/src/main/java/com/softicar/platform/emf/management/EmfManagementDiv.java
```diff
@@ -186,20 +188,39 @@ public ActionBar() {
...
-			appendChild(
-				new DomPopupButton()//
-					.setPopupFactory(() -> new EmfImportPopup<>(entityTable, scopeEntity))
-					.setIcon(EmfImages.ENTITY_IMPORT.getResource())
-					.setLabel(EmfI18n.IMPORT)
-					.setMarker(EmfManagementMarker.IMPORT_BUTTON)
-					.setEnabled(isCreationAllowed())
-					.setTitle(getCreationPredicateTitle()));
-
+			if (!isNonConcealedNonNullableForeignAttributePresent()) {
```
oliver-richers, line 191: hmm, not sure why we need to disable the button in this case, but, anyway

oliver-richers (approved):
I am not very glad with these changes, but ok, it does not hurt for now

PR #149, opened 2022-02-16
PLAT-699 Update tab title when changing pages

daniel-klose:
image

### DIFF
--- a/platform-ajax/src/main/java/com/softicar/platform/ajax/engine/AjaxDomEngine.java
+++ b/platform-ajax/src/main/java/com/softicar/platform/ajax/engine/AjaxDomEngine.java
@@ -492,7 +492,7 @@ public class AjaxDomEngine implements IDomEngine {
 	@Override
 	public void updateDocumentTitle(String pageTitle) {
 
-		updateCodeJS.appendStatement("document.title = \"" + pageTitle + "\";");
+		JS_setAttribute("document.title", pageTitle);
 	}
 
 	@Override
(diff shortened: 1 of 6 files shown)


platform-ajax/src/main/java/com/softicar/platform/ajax/engine/AjaxDomEngine.java
```diff
@@ -489,6 +489,12 @@ public AjaxResourceLinkRegistry getResourceLinkRegistry() {
 
 	// -------------------------------- browser manipulation -------------------------------- //
 
+	@Override
+	public void updateDocumentTitle(String pageTitle) {
```
oliver-richers, line 493: why not `setDocumentTitle(IDisplayString)`? wouldn't that be better

platform-ajax/src/main/java/com/softicar/platform/ajax/engine/AjaxDomEngine.java
```diff
@@ -536,6 +542,11 @@ private void JS_removeNodeAttribute(IDomNode node, String name) {
 		updateCodeJS.appendStatement(JS_getNode(node) + ".removeAttribute('" + name + "');");
 	}
 
+	private void JS_setAttribute(String attribute, String value) {
```
oliver-richers, line 545: hmm, isn't this method redundant? shouldn't be the first occasion where we set an attribute

oliver-richers (changes requested):

platform-ajax/src/main/java/com/softicar/platform/ajax/engine/AjaxDomEngine.java
daniel-klose: dunno, changed to set

platform-ajax/src/main/java/com/softicar/platform/ajax/engine/AjaxDomEngine.java
daniel-klose: I did not find any such method, please give an example :)

oliver-richers: ok

oliver-richers (approved):

PR #153, opened 2022-02-22
PLAT-687 Add more Items indicator to AutoCompleteInputs

daniel-klose:
image
image

### DIFF
--- a/platform-ajax/src/main/java/com/softicar/platform/ajax/document/creation/AjaxDocumentPagePrinter.java
+++ b/platform-ajax/src/main/java/com/softicar/platform/ajax/document/creation/AjaxDocumentPagePrinter.java
@@ -86,6 +86,7 @@ class AjaxDocumentPagePrinter {
 		// texts
 		println("var LOCK_MESSAGE = '%s';", AjaxI18n.PLEASE_WAIT.concat(" ").concat(AjaxI18n.ANOTHER_ACTION_IS_STILL_EXECUTING));
 		println("var AUTO_COMPLETE_TEXT_GENERIC = '%s';", AjaxI18n.AUTO_COMPLETE_INPUT_ELEMENT);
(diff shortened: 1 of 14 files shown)


platform-dom/src/main/java/com/softicar/platform/dom/input/auto/DomAutoCompleteList.java
```diff
@@ -12,7 +12,7 @@
  */
 public class DomAutoCompleteList implements Iterable<String> {
 
-	public static final int MAXIMUM_ELEMENT_COUNT = 16;
+	public static final int MAXIMUM_ELEMENT_COUNT = 17;
```
oliver-richers, line 15: generally, this seems to be ok, although the fact the here we have the constant 17 and in JS, we have 16 will be confusing for anybody how looks at it in the future

oliver-richers (changes requested):
we should somehow improve this --> discuss with Alex

platform-ajax/src/main/java/com/softicar/platform/ajax/document/creation/AjaxDocumentPagePrinter.java
```diff
@@ -86,6 +86,7 @@ private void printJavascriptFramework() {
 		// texts
 		println("var LOCK_MESSAGE = '%s';", AjaxI18n.PLEASE_WAIT.concat(" ").concat(AjaxI18n.ANOTHER_ACTION_IS_STILL_EXECUTING));
 		println("var AUTO_COMPLETE_TEXT_GENERIC = '%s';", AjaxI18n.AUTO_COMPLETE_INPUT_ELEMENT);
+		println("var AUTO_COMPLETE_TEXT_FURTHER_ENTRIES_AVAILABLE = '%s';", AjaxI18n.FURTHER_ENTRIES_AVAILABLE);
```
daniel-klose, line 89: Now that I think about this:
We could have just defined another new variable here which represents `DomAutoCompleteList.MAXIMUM_ELEMENT_COUNT` and use it in the javascript.
With that, there would be way less code to adapt than the current approach.

PR #156, opened 2022-02-23
PLAT-687 Add more Items info to AutoCompleteInputs

daniel-klose:
image
image

### DIFF
--- a/platform-dom/src/main/java/com/softicar/platform/dom/input/auto/DomAutoCompleteList.java
+++ b/platform-dom/src/main/java/com/softicar/platform/dom/input/auto/DomAutoCompleteList.java
@@ -12,11 +12,8 @@ import java.util.List;
  */
 public class DomAutoCompleteList implements Iterable<String> {
 
-	// The maximum amount of elements that are displayed
-	public static final int MAXIMUM_ELEMENT_COUNT = 16;
-
-	// The maximum amount of rows that are loaded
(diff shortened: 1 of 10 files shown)


platform-dom/src/main/java/com/softicar/platform/dom/input/auto/DomAutoCompleteList.java
```diff
@@ -13,6 +13,7 @@
 public class DomAutoCompleteList implements Iterable<String> {
```
oliver-richers, line 13: hmm, the constants deserve a comment, or better, make them both private and provide a static getter method with Javadoc, probably only for `MAXIMUM_ROW_COUNT`

oliver-richers (approved):
except from my comment the PR looks fine

platform-dom/src/main/java/com/softicar/platform/dom/input/auto/DomAutoCompleteList.java
oliver-richers: I find the distinction between ROW and ELEMENT quite esoteric, why not make the constant names speak for themselves? e.g. `MAXIMUM_ELEMENT_COUNT` -> `MAXIMUM_ELEMENTS_TO_DISPLAY` and `MAXIMUM_ROW_COUNT` -> `MAXIMUM_ELEMENTS_TO_LOAD`

oliver-richers (approved):
looks fine now

PR #164, opened 2022-03-03
PLAT-601 Defined ILocale, CurrentLocale, etc.

oliver-richers:


platform-common-core/src/main/java/com/softicar/platform/common/core/locale/LocaleValidator.java
```diff
@@ -0,0 +1,25 @@
...
+
+	public void validate() {
+
+		if (locale.getDecimalSeparator().isEmpty()) {
+			throw new SofticarUserException(CommonCoreI18n.THE_DECIMAL_SEPARATOR_MAY_NOT_BE_EMPTY);
+		}
+
+		if (locale.getDecimalSeparator().equals(locale.getDigitGroupSeparator())) {
+			throw new SofticarUserException(CommonCoreI18n.THE_DECIMAL_SEPARATOR_MUST_BE_DIFFERENT_FROM_THE_DIGIT_GROUP_SEPARATOR);
+		}
```
daniel-klose, line 23: I think there needs to be an additional test that the Seperators may only be a single symbol.
Also, if easily implemented, there should be a check that the separators may not be letters like `c` or numbers `5`

daniel-klose (changes requested):

platform-common-core/src/main/java/com/softicar/platform/common/core/locale/LocaleValidator.java
oliver-richers: - hmm, does it have to be a single character? and why prevent letters?
- regardings digits, you are right, will add that

daniel-klose: well tbh, everything beside a single special character is pretty much useless as a seperator imo ¯\_(ツ)_/¯

alexander-schmidt: I think:
- Most characters might be valid separators - somewhere, for some use case. Even a space. Even emoji. Notable exceptions are **digits** and **control characters**. We _could_ enforce that those are not used - but it's probably not necessary as long as end-users cannot freely define individual locales. If an admin deliberately _wants_ to break the configuration by defining something dumb - let them.
- Allowing multi-character separators is likely to cause headache in all sorts of parsers/converters. We should avoid that unless absolutely necessary.

platform-common-core/src/main/java/com/softicar/platform/common/core/locale/ILocale.java
```diff
@@ -0,0 +1,34 @@
...
+	 * Returns the {@link String} to be used as decimal separator.
+	 * <p>
+	 * Common values are the decimal comma (,) or decimal point (.).
+	 * <p>
+	 * This separator must be different to the separator returned by
+	 * {@link #getDigitGroupSeparator()}.
+	 *
+	 * @return the decimal separator (never <i>null</i> and not empty)
+	 */
+	String getDecimalSeparator();
```
alexander-schmidt, line 20: We could change this to return `Optional<Character>`. AFAIK, multi-character separators are not a thing.

platform-common-core/src/main/java/com/softicar/platform/common/core/locale/ILocale.java
```diff
@@ -0,0 +1,34 @@
...
+	 * <p>
+	 * Common values are space ( ), comma (,), point (.) or apostrophe (') or
+	 * the empty {@link String}.
+	 * <p>
+	 * This separator must be different to the separator returned by
+	 * {@link #getDecimalSeparator()}.
+	 *
+	 * @return the decimal separator (never <i>null</i> but may be empty)
+	 */
+	String getDigitGroupSeparator();
```
alexander-schmidt, line 33: We could change this to return `Optional<Character>`. AFAIK, multi-character separators are not a thing.

platform-common-core/src/main/java/com/softicar/platform/common/core/locale/ILocale.java
oliver-richers: I had this previously, but I prefer String here.

platform-common-core/src/main/java/com/softicar/platform/common/core/locale/ILocale.java
oliver-richers: No, I think String is better here. IMO, Optional<Character> would be overengineering.

platform-common-core/src/main/java/com/softicar/platform/common/core/locale/LocaleValidator.java
oliver-richers: I don't see a problem with multi character separators, but I can add this for now.

oliver-richers: On the other hand, no, I would only do this if we really run into problems.

daniel-klose (approved):

PR #167, opened 2022-03-04
PLAT-601-2 Implemented BigDecimalFormatter

oliver-richers:


platform-common-core/src/test/java/com/softicar/platform/common/core/number/formatter/BigDecimalFormatterTest.java
```diff
@@ -0,0 +1,77 @@
+package com.softicar.platform.common.core.number.formatter;
+
+import com.softicar.platform.common.core.locale.ILocale;
+import com.softicar.platform.common.core.locale.Locale;
+import java.math.BigDecimal;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class BigDecimalFormatterTest extends Assert {
```
alexander-schmidt, line 9: This test does not, in any way, ensure that `BigDecimalFormatter` uses `CurrentLocale`. However, this is likely to be the most common use case.

We could even manipulate `BigDecimalFormatter`, replacing
`this.locale = CurrentLocale.get();`
with
`this.locale = null`
The thing would crash (unless `setLocale` gets called) - yet all tests will still be green.

alexander-schmidt (changes requested):

platform-common-core/src/test/java/com/softicar/platform/common/core/number/formatter/BigDecimalFormatterTest.java
oliver-richers: did so; to made this testable, I introduced new `LocaleScope`

platform-common-core/src/main/java/com/softicar/platform/common/core/locale/LocaleScope.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+public class LocaleScope implements AutoCloseable {
+
+	private final ILocale originalLocale;
+
+	/**
+	 * Sets {@link CurrentLocale} to the given {@link ILocale}.
+	 *
+	 * @param locale
+	 *            the language to set (never null)
```
alexander-schmidt, line 16:
```suggestion
	 *            the language to set (never <i>null</i>)
```

alexander-schmidt (changes requested):

platform-common-core/src/main/java/com/softicar/platform/common/core/locale/LocaleScope.java
oliver-richers: fixed, also fixed `language` -> `ILocale`

alexander-schmidt (approved):

PR #169, opened 2022-03-07
PLAT-724 Fix Style of LimitedWidthZoomableImage

daniel-klose:
Also refactored stuff to use CSS classes
image
Note that the arrows and page count are now centered


oliver-richers (approved):
looks fine

PR #171, opened 2022-03-08
PLAT-718 Don't Show Messages and StackTrace of Non-SofticarUserExceptions

thees-k:
### Summary
In case of an unexpected error, the exception message will not be shown anymore. Instead a standard message will be shown: "An internal program error occurred."
### Test Plan
* Provoke an exception that is not a `SofticarUserException` e.g. by putting too many characters inside an input field and then click "Save":

* Check if the exception was stored properly:

* Also note the new test method.


platform-ajax/src/main/java/com/softicar/platform/ajax/document/service/AbstractAjaxDocumentService.java
```diff
@@ -45,11 +47,21 @@ protected void executePayloadCode(INullaryVoidFunction payloadCode) {
 		try {
 			payloadCode.apply();
 		} catch (Exception exception) {
-			document.getBody().executeAlert(IDisplayString.create(exception.getLocalizedMessage()));
+			IDisplayString message = getDisplayMessage(exception);
+			document.getBody().executeAlert(message);
```
oliver-richers, line 51: we could inline this 
```suggestion
			document.getBody().executeAlert(getDisplayMessage(exception));
```

platform-ajax/src/test/java/com/softicar/platform/ajax/event/AjaxDomEventWithExceptionTest.java
```diff
@@ -44,4 +46,18 @@ public void test() {
 		// assert text that was appended before the Exception occurred is displayed
 		assertEquals("Hello", getText(output));
 	}
+
+	@Test
+	public void testWithNonSofticarUserException() {
+
```
oliver-richers, line 52: should we also test that the document was properly created?
```suggestion
// check that the document was created properly
assertEquals("", getText(output));
```

platform-ajax/src/test/java/com/softicar/platform/ajax/event/AjaxDomEventWithExceptionTest.java
```diff
@@ -44,4 +46,18 @@ public void test() {
...
+		// click exception-throwing button
+		click(nonSofticarUserExceptionButton);
+		waitForServer();
+
+		// assert modal alert is displayed
+		var alert = findModalAlertOrFail();
+		assertFocused(DomModalAlertMarker.CLOSE_BUTTON);
+		assertEquals("An internal program error occurred.", getText(alert.getContent()));
```
oliver-richers, line 60: I think we should use `AjaxI18n.AN_INTERNAL_PROGRAM_ERROR_OCCURRED` instead of a literal string

oliver-richers (changes requested):

oliver-richers (approved):

PR #173, opened 2022-03-08
PLAT-601-4 Implemented BigDecimalParser and Strings

oliver-richers:


platform-common-core/src/main/java/com/softicar/platform/common/core/number/parser/BigDecimalParser.java
```diff
@@ -0,0 +1,118 @@
...
+			var integralPart = numberString.substring(0, decimalSeparatorIndex);
+			var fractionalPart = numberString.substring(decimalSeparatorIndex + decimalSeparator.length());
+			validateIntegralPart(integralPart);
+			validateFractionalPart(fractionalPart);
+		}
+	}
+
+	private void validateIntegralPart(String integralPart) {
+
+		String[] groups = integralPart.replace(locale.getDigitGroupSeparator(), ",").split(",");
```
thees-k, line 99: Why not just
`String[] groups = integralPart.split(locale.getDigitGroupSeparator());`
?

platform-common-core/src/main/java/com/softicar/platform/common/core/number/parser/BigDecimalParser.java
```diff
@@ -0,0 +1,118 @@
...
+			validateIntegralPart(numberString);
+		} else {
+			var integralPart = numberString.substring(0, decimalSeparatorIndex);
+			var fractionalPart = numberString.substring(decimalSeparatorIndex + decimalSeparator.length());
+			validateIntegralPart(integralPart);
+			validateFractionalPart(fractionalPart);
+		}
+	}
+
+	private void validateIntegralPart(String integralPart) {
```
thees-k, line 97: The integral part may consist of digits and digitGroupSeparator only. No other characters are allowed. I think this should be validated.

platform-common-core/src/main/java/com/softicar/platform/common/core/number/parser/BigDecimalParser.java
```diff
@@ -0,0 +1,118 @@
...
+					throw new SofticarUserException(CommonCoreI18n.DIGIT_GROUP_TOO_LONG);
+				}
+				if (i > 0 && groups[i].length() < DIGIT_GROUP_SIZE) {
+					throw new SofticarUserException(CommonCoreI18n.DIGIT_GROUP_TOO_SHORT);
+				}
+			}
+		}
+	}
+
+	private void validateFractionalPart(String fractionalPart) {
```
thees-k, line 112: The fractional part may consist of digits and decimalSeparator only. No other characters are allowed. I think this should be validated.

thees-k (changes requested):
I think the methods `validateIntegralPart()` and `validateFractionalPart()` should be improved (please see the commented lines).

platform-common-core/src/main/java/com/softicar/platform/common/core/number/parser/BigDecimalParser.java
oliver-richers: new later `new BigDecimal(numberString)` call will already catch that

thees-k (approved):

PR #178, opened 2022-03-09
PLAT-145 Rework DomDayPopup to be a DomDayPopover

daniel-klose:
image
  - Removed some unused code


platform-dom/src/main/java/com/softicar/platform/dom/elements/time/day/DomDayChooserDiv.java
```diff
@@ -267,7 +253,10 @@ public TodayButton() {
 
 			setIcon(DomElementsImages.CALENDAR_TODAY.getResource());
 			setLabel(DomI18n.TODAY);
-			setClickCallback(() -> setDay(Day.today()));
+			setClickCallback(() -> {
```
alexander-schmidt, line 256: We should probably extract a private method here, instead of using a block.

platform-dom/src/main/java/com/softicar/platform/dom/elements/time/day/AbstractDomDayPopupButton.java
```diff
@@ -9,34 +9,28 @@
...
- * change the date and close the pop-up menuToClose the changes you have made
- * will be shown.
+ * An object of this class already contains a {@link DomDayPopover} object. If
+ * you change the date, the popover closes and the changes will be shown.
  *
  * @author Oliver Richers
  * @author Alexander Schmidt
  */
-public abstract class AbstractDomDayPopupButton extends DomButton implements IDomInputNode {
+abstract class AbstractDomDayPopupButton extends DomButton implements IDomInputNode {
```
alexander-schmidt, line 18: I think the behavoir of this thing is great now! We should ensure that is stays like this, by adding test cases to `DomDayInputTest` (particularly, tests covering the popup).

alexander-schmidt (changes requested):

platform-dom/src/test/java/com/softicar/platform/dom/elements/time/day/DomDayPopoverTest.java
```diff
@@ -0,0 +1,76 @@
...
+import com.softicar.platform.dom.elements.button.DomButton;
+import com.softicar.platform.dom.elements.popup.modal.DomPopover;
+import com.softicar.platform.dom.elements.testing.engine.IDomTestEngine;
+import com.softicar.platform.dom.elements.testing.engine.IDomTestEngineMethods;
+import com.softicar.platform.dom.elements.testing.engine.document.DomDocumentTestEngine;
+import com.softicar.platform.dom.input.DomTextInput;
+import org.junit.Before;
+import org.junit.Test;
+
+public class DomDayPopoverTest extends AbstractTest implements IDomTestEngineMethods {
```
alexander-schmidt, line 18: I think we should add the following test cases:
- What day is highlighted in the popover if the button is clicked while the input is empty?
- What day is highlighted in the popover if the button is clicked while the input contains garbage?
- What day is highlighted in the popover if the button is clicked while the input contains a valid date?
- What happens upon ESC?
- What happens if the user selects a different year or month (and therefore enforces that the previously-selected day is no longer selectable), and then dismisses the popover?

alexander-schmidt (changes requested):

alexander-schmidt (approved):

PR #179, opened 2022-03-14
PLAT-601-6 Unified Inputs for BigDecimal, Double and Float

oliver-richers:


platform-common-core/src/main/java/com/softicar/platform/common/core/number/BigDecimalMapper.java
```diff
@@ -0,0 +1,35 @@
+package com.softicar.platform.common.core.number;
+
+import java.math.BigDecimal;
+import java.util.function.Function;
+
+/**
+ * A class to map between {@link Number} and {@link BigDecimal}.
```
alexander-schmidt, line 7:
```suggestion
 * A class to map between {@link BigDecimal} and a {@link Number} based type.
```

platform-common-core/src/main/java/com/softicar/platform/common/core/number/formatter/BigDecimalScaleApplier.java
```diff
@@ -0,0 +1,92 @@
+package com.softicar.platform.common.core.number.formatter;
+
+import java.math.BigDecimal;
+import java.math.RoundingMode;
+
+/**
+ * Applies a given scale to {@link BigDecimal} values.
```
alexander-schmidt, line 7:
```suggestion
 * Applies a given scale to a {@link BigDecimal} value.
```
...since this whole class is built around modifications of a single value.

platform-common-core/src/main/java/com/softicar/platform/common/core/number/formatter/BigDecimalScaleApplier.java
```diff
@@ -0,0 +1,92 @@
...
+
+	/**
+	 * Defines the desired scale of the values, i.e. the number of decimal
+	 * places.
+	 * <p>
+	 * If <i>null</i> is provided as scale (which is the default), no changes
+	 * will be applied to the value.
+	 *
+	 * @param scale
+	 *            the scale; or <i>null</i> to disable scaling
```
alexander-schmidt, line 30: "disable scaling" - what does that mean? Seems to imply that a previously-scaled value could become unscaled if `null` is given here.
suggestions:
- `to perform no scaling`
- `to omit scaling`
- `to skip scaling`
- `to not apply scaling`
- `to not change scaling`
- ...

platform-common-core/src/main/java/com/softicar/platform/common/core/number/formatter/BigDecimalScaleApplier.java
```diff
@@ -0,0 +1,92 @@
...
+	 * @return this
+	 */
+	public BigDecimalScaleApplier setScale(Integer scale) {
+
+		this.scale = scale;
+		return this;
+	}
+
+	/**
+	 * Defines the {@link RoundingMode} for the values.
```
alexander-schmidt, line 40:
```suggestion
	 * Defines the {@link RoundingMode} for the value.
```

platform-common-core/src/main/java/com/softicar/platform/common/core/number/formatter/BigDecimalScaleApplier.java
```diff
@@ -0,0 +1,92 @@
...
+	 * <p>
+	 * If <i>null</i> is provided as {@link RoundingMode} (which is the
+	 * default), no rounding will be done. Only redundant, trailing zeros will
+	 * be added or discarded from the fractional part to approximate the desired
+	 * scale as possible.
+	 *
+	 * @param roundingMode
+	 *            the {@link RoundingMode}; or <i>null</i> to disable rounding
```
alexander-schmidt, line 48: see above, regarding "disable"

platform-common-core/src/main/java/com/softicar/platform/common/core/number/formatter/BigDecimalScaleApplier.java
```diff
@@ -0,0 +1,92 @@
...
+	 * @return this
+	 */
+	public BigDecimalScaleApplier setRoundingMode(RoundingMode roundingMode) {
+
+		this.roundingMode = roundingMode;
+		return this;
+	}
+
+	/**
+	 * Applies the scale defined by calling {@link #setScale(Integer)}.
```
alexander-schmidt, line 58: I think this Javadoc should at least mention "rounding" or "rounding mode".

platform-dom/src/main/java/com/softicar/platform/dom/elements/number/decimal/DomDecimalInput.java
```diff
@@ -0,0 +1,86 @@
...
+		return new BigDecimalFormatter(decimal).format();
+	}
+
+	@Override
+	protected T parseValue(String text) {
+
+		try {
+			var decimal = getScaled(new BigDecimalParser(text).parseOrThrow());
+			if (scale >= 0 && decimal.scale() > scale) {
+				throw new DomInputException(DomI18n.ONLY_ARG1_DECIMAL_PLACES_ALLOWED.toDisplay(scale));
```
alexander-schmidt, line 68: The message seems to imply an expected _exact_ number of decimal places. Should contain "max", "at most", or sth similar.

platform-dom/src/main/java/com/softicar/platform/dom/elements/number/decimal/DomDecimalDisplay.java
```diff
@@ -0,0 +1,54 @@
...
+ * @author Oliver Richers
+ */
+public class DomDecimalDisplay<T extends Number> extends DomDiv {
+
+	private final BigDecimalMapper<T> mapper;
+	private Integer scale;
+
+	public DomDecimalDisplay(BigDecimalMapper<T> mapper) {
+
+		this.mapper = mapper;
```
alexander-schmidt, line 25:
```suggestion
		this.mapper = Objects.requireNonNull(mapper);
```
Let's fail early here.

platform-dom/src/main/java/com/softicar/platform/dom/elements/number/decimal/DomDecimalDisplay.java
```diff
@@ -0,0 +1,54 @@
...
+		this.scale = null;
+	}
+
+	public DomDecimalDisplay<T> setScale(int scale) {
```
alexander-schmidt, line 29: I wonder if the use of `int` here is intentional:
`this.scale` defaults to `null`, so this class can indeed handle that case. Particularly, because we tolerate a `null` scale in `BigDecimalScaleApplier`.
Should the parameter be changed to `Integer`?

platform-dom/src/main/java/com/softicar/platform/dom/elements/number/decimal/DomDecimalDisplay.java
```diff
@@ -0,0 +1,54 @@
...
+import java.math.BigDecimal;
+import java.math.RoundingMode;
+
+/**
+ * An {@link IDomElement} to display decimal values according to
+ * {@link CurrentLocale}.
+ *
+ * @author Oliver Richers
+ */
+public class DomDecimalDisplay<T extends Number> extends DomDiv {
```
alexander-schmidt, line 18: I think this element deserves its own unit test, considering its importance: We'll use it to display values all over the place, in ERP software. So we should ensure correct rounding/scaling/truncation/whatever behavior.

platform-dom/src/main/java/com/softicar/platform/dom/elements/number/decimal/DomDecimalInput.java
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * A {@link DomTextInput} for decimal values that formats and parses values
+ * according to {@link CurrentLocale}.
+ *
+ * @author Oliver Richers
+ */
+public class DomDecimalInput<T extends Number> extends AbstractDomNumberInput<T> {
+
+	private final BigDecimalMapper<T> mapper;
+	private int scale = -1;
```
alexander-schmidt, line 23: I think this is confusing, and that we should rather use a `null` `Integer` to indicate an undefined scale.
After all, even though rarely used, negative scales are a thing, in a `BigDecimal` context. Yet, the value of `-1` here is not actually meant to be a negative scale - it's merely used to not have to bother with `null` checks.

platform-dom/src/main/java/com/softicar/platform/dom/elements/number/decimal/DomDecimalInput.java
```diff
@@ -0,0 +1,86 @@
...
+	 * exception.
+	 * <p>
+	 * By default, i.e. when this method was not called, this input does not
+	 * modify the scale of values in this input.
+	 *
+	 * @param scale
+	 *            the number of decimal places which must be <i>>= 0</i>
+	 * @return this
+	 */
+	public DomDecimalInput<T> setScale(int scale) {
```
alexander-schmidt, line 46: I think we should add `@throws` here, e.g.
`@throws IllegalArgumentException if a negative value is given`

platform-emf/src/main/java/com/softicar/platform/emf/attribute/field/decimal/EmfDecimalAttribute.java
```diff
@@ -0,0 +1,73 @@
...
+		this.mapper = mapper;
+		this.scale = Optional.empty();
+
+		setDisplayFactory(Display::new);
+		setInputFactory(Input::new);
+	}
+
+	public EmfDecimalAttribute<R, V> setScale(int scale) {
+
+		this.scale = Optional.of(scale);
```
alexander-schmidt, line 34: `Optional.of(<int>)` is redundant.
Also, we should _probably_ change the parameter type to `Integer`, and tolerate null -- see related comments on `DomDecimalDisplay`.

alexander-schmidt (changes requested):

platform-common-core/src/main/java/com/softicar/platform/common/core/number/BigDecimalMapper.java
oliver-richers: ok

platform-common-core/src/main/java/com/softicar/platform/common/core/number/formatter/BigDecimalScaleApplier.java
oliver-richers: actually, in this case, the class instance can be used multiple times, i wonder if I should change BigDecimalFormatter in a similar way, so it can be configured an used for multiple values?

platform-common-core/src/main/java/com/softicar/platform/common/core/number/formatter/BigDecimalScaleApplier.java
```diff
@@ -0,0 +1,92 @@
...
+	private RoundingMode roundingMode;
+
+	public BigDecimalScaleApplier() {
+
+		this.scale = null;
+		this.roundingMode = null;
+	}
+
+	/**
+	 * Defines the desired scale of the values, i.e. the number of decimal
```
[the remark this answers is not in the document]
oliver-richers: see above

platform-common-core/src/main/java/com/softicar/platform/common/core/number/formatter/BigDecimalScaleApplier.java
oliver-richers: ok

platform-common-core/src/main/java/com/softicar/platform/common/core/number/formatter/BigDecimalScaleApplier.java
oliver-richers: see above

platform-common-core/src/main/java/com/softicar/platform/common/core/number/formatter/BigDecimalScaleApplier.java
oliver-richers: ok

platform-common-core/src/main/java/com/softicar/platform/common/core/number/formatter/BigDecimalScaleApplier.java
oliver-richers: did so

platform-dom/src/main/java/com/softicar/platform/dom/elements/number/decimal/DomDecimalInput.java
oliver-richers: changed to `No more than %s decimal places allowed.`

platform-dom/src/main/java/com/softicar/platform/dom/elements/number/decimal/DomDecimalDisplay.java
oliver-richers: did so

platform-dom/src/main/java/com/softicar/platform/dom/elements/number/decimal/DomDecimalDisplay.java
oliver-richers: did so and added extensive Javadoc

platform-dom/src/main/java/com/softicar/platform/dom/elements/number/decimal/DomDecimalInput.java
oliver-richers: changed this, even allowing negative scale, although it will have no effect

platform-dom/src/main/java/com/softicar/platform/dom/elements/number/decimal/DomDecimalInput.java
oliver-richers: well this is obsolete now, see above

platform-emf/src/main/java/com/softicar/platform/emf/attribute/field/decimal/EmfDecimalAttribute.java
oliver-richers: i agree, since this also simplifies code; changed to `Integer` and added Javadoc

platform-dom/src/main/java/com/softicar/platform/dom/elements/number/decimal/DomDecimalDisplay.java
oliver-richers: added DomDecimalDisplayTest **and** BigDecimalScaleApplierTest

platform-common-core/src/main/java/com/softicar/platform/common/core/number/formatter/BigDecimalScaleApplier.java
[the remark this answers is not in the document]
oliver-richers: rephrased

platform-common-core/src/main/java/com/softicar/platform/common/core/number/formatter/BigDecimalScaleApplier.java
oliver-richers: rephrased

platform-common-core/src/main/java/com/softicar/platform/common/core/number/formatter/BigDecimalScaleApplier.java
oliver-richers: - not sure if the exact wording is so important here, but i changed it, as you wished
- changed BigDecimalFormatter

alexander-schmidt (approved):

PR #184, opened 2022-03-17
PLAT-741 Data Tables: Use Auto-Complete inputs in entity filters

alexander-schmidt:
Peek 2022-03-17 09-47


oliver-richers (approved):
nice

PR #188, opened 2022-03-17
PLAT-736 Added IDomEnableable.isDisabled() and IDomTextualInput.isReadonly()

oliver-richers:
Test Plan:
- start TestCasesDiv
- click on the test buttons to disable/enable the input
- click on the test buttons to readonly/read-write the input

image


platform-dom/src/main/java/com/softicar/platform/dom/input/IDomTextualInput.java
```diff
@@ -88,4 +88,28 @@ default void select() {
...
 		getDomEngine().select(this);
 	}
+
+	// -------------------------------- read-only -------------------------------- //
+
+	/**
+	 * Makes this {@link IDomTextualInput} read-only.
+	 * <p>
+	 * In contrast to {@link #setEnabled(boolean)}, the user can still see and
+	 * copy the text of this {@link IDomTextualInput}.
```
thees-k, line 98: I tested with Firefox. You _can_ copy the text of a disabled input field, too. So this Javadoc comment is not right.

thees-k (changes requested):
I tested `TestCasesDiv` with Firefox.

The only difference between "read only" and "disable" seems to be that the text of a disabled input field is much more transparent.

Maybe we don't need "read only" at all?

platform-dom/src/main/java/com/softicar/platform/dom/input/IDomTextualInput.java
oliver-richers: interesting

oliver-richers: I removed this comment.

thees-k (approved):

PR #195, opened 2022-03-23
PLAT-744 Fix minor database issues

alexander-schmidt:
- `AGCoreModuleInstance`: Changed `defaultLocalization` to non-null.
- `IDbEnumTableRowEnum`: No longer try to load enum table records, to avoid problems with DB initialization order.

Unrelated:
- Added some TODOs.
- Removed some redundant whitespaces.


oliver-richers (approved):
ok

PR #199, opened 2022-03-28
PLAT-733 IEmfInput: Add setValueAndHandleChangeCallback

daniel-klose:
- `setValueAndHandleChangeCallback` in `IEmfInput` manually executes the change-callback, if present
  - previously, if setting a value of an input with just `setValue`, no change-callback was executed, which lead to some problems
  - added a test implementation in `AGDemoMoment` where a button sets the value of the input with the new method
  - added a dependee of the field that gets its value set programmatically that updates its value via `refreshInputConstraints`
image
image
image

### DIFF
--- a/platform-demo-module/src/main/java/com/softicar/platform/demo/module/moment/DemoMomentPointInTimeInput.java
+++ b/platform-demo-module/src/main/java/com/softicar/platform/demo/module/moment/DemoMomentPointInTimeInput.java
@@ -1,0 +1,21 @@
+package com.softicar.platform.demo.module.moment;
+
+import com.softicar.platform.emf.attribute.field.daytime.EmfDayTimeInput;
+
+public class DemoMomentPointInTimeInput extends EmfDayTimeInput {
+
+	private final AGDemoMoment moment;
+
+	public DemoMomentPointInTimeInput(AGDemoMoment moment) {
+
+		this.moment = moment;
+	}
+
+	@Override
+	public void refreshInputConstraints() {
+
+		if (moment.getDay() != null) {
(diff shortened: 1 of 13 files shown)


platform-emf/src/main/java/com/softicar/platform/emf/attribute/field/bool/EmfBooleanInput.java
```diff
@@ -30,6 +30,13 @@ public void setValue(Boolean value) {
 		checkBox.setChecked(value);
 	}
 
+	@Override
+	public void setValueAndHandleChangeCallback(Boolean value) {
+
+		// DomCheckbox#setChecked already executes a callback
```
oliver-richers, line 36: uuuh, we should reconsider that, maybe add reference to PLAT-756

platform-demo-module/src/main/java/com/softicar/platform/demo/module/moment/DemoMomentPointInTimeInput.java
```diff
@@ -0,0 +1,21 @@
...
+
+	private final AGDemoMoment moment;
+
+	public DemoMomentPointInTimeInput(AGDemoMoment moment) {
+
+		this.moment = moment;
+	}
+
+	@Override
+	public void refreshInputConstraints() {
```
oliver-richers, line 15: - not sure if this test shows a realistic example, since changing the input value on `refreshInputConstraints` is quite unexpected, any ideas?

oliver-richers (changes requested):

oliver-richers (approved):

PR #204, opened 2022-03-31
PLAT-764 EmfImport: Improved Naming of Variables of CSV Files

thees-k:


platform-emf/src/main/java/com/softicar/platform/emf/management/importing/variable/find/EmfImportVariablesFinder.java
```diff
@@ -6,7 +6,7 @@
 
 public class EmfImportVariablesFinder {
 
-	private static final String VARIABLE_PATTERN = "\\$\\$[a-z]\\$\\$";
+	private static final String VARIABLE_PATTERN = "\\$\\$[a-z]+\\$\\$";
```
alexander-schmidt, line 9: I think we should allow `[a-zA-Z0-9_]` ... and probably even spaces.
Being less restrictive here will result in improved legibility of multi-word variable names. (e.g. `businessPartnerName` or `business_partner_name` instead of `businesspartnername`)

oliver-richers, line 9: i think we should also support underscore and uppercase letters as well as numbers

alexander-schmidt (changes requested):

oliver-richers (changes requested):

platform-emf/src/main/java/com/softicar/platform/emf/management/importing/variable/find/EmfImportVariablesFinder.java
thees-k: See new ACCs of story

thees-k: See new ACCs of story.

PR #214, opened 2022-04-11
PLAT-767 Collapse Navi-Folders is now configurable

daniel-klose:


platform-core-module/src/main/java/com/softicar/platform/core/module/CoreI18n.java
```diff
@@ -130,6 +130,8 @@ public interface CoreI18n extends EmfI18n {
 		.de("Klicken, um die Programm-Logs anzuzeigen.");
 	I18n0 CLIENT_IP_ADDRESS = new I18n0("Client IP Address")//
 		.de("Client IP Adresse");
+	I18n0 COLLAPSE_NAVIGATION_FOLDERS = new I18n0("Collapse Navigation Folders")//
```
alexander-schmidt, line 133:
```suggestion
	I18n0 AUTOMATICALLY_COLLAPSE_FOLDERS = new I18n0("Automatically Collapse Folders")//
```
Since this is displayed in a "Navigation"-titled fieldset, we could omit "Navigation" from these identifiers... kind of redundant. Also, I think that adding "Automatically" adds some clarity here.
(DB field name could remain as-is, btw)

platform-core-module/src/main/java/com/softicar/platform/core/module/CoreI18n.java
```diff
@@ -130,6 +130,8 @@ public interface CoreI18n extends EmfI18n {
...
+		.de("Navigationsordner einklappen");
```
alexander-schmidt, line 134:
```suggestion
		.de("Ordner automatisch einklappen");
```

platform-core-module/src/test/java/com/softicar/platform/core/module/page/PageDivTest.java
```diff
@@ -85,8 +85,21 @@ public void testOpeningPageWithParameters() {
 	}
 
 	@Test
-	public void testMultipleOpenFolder() {
+	public void testMultipleOpenFolderWithCollapsing() {
```
alexander-schmidt, line 89: For some extra clarity, we could add `testUser.setCollapseNavigationFolders(true).save();` here

platform-integration/src/main/java/com/softicar/platform/integration/database/structure/DatabaseStructureJsonFromClasspathExtractor.java
```diff
@@ -17,6 +17,11 @@ public class DatabaseStructureJsonFromClasspathExtractor {
 
 	private final String tablePackagePrefix;
 
+//	public static void main(String[] args) {
```
alexander-schmidt, line 20: We could add a comment above this comment. Like "Use this main method to blablabla"

alexander-schmidt (changes requested):

platform-core-module/src/main/java/com/softicar/platform/core/module/CoreI18n.java
daniel-klose: Db name needs to change aswell since the auto generation forces this

alexander-schmidt (approved):

PR #229, opened 2022-04-26
PLAT-728-1 Removed IDomInputNode and Improved Input Interfaces

oliver-richers:
IDomDisableable is now only impemented by nodes that can actually be disabled:
image
Same for IDomFocusable:
image


platform-dom/src/main/java/com/softicar/platform/dom/elements/DomLabel.java
```diff
@@ -2,18 +2,16 @@
...
 import com.softicar.platform.dom.parent.DomParentElement;
 
 /**
- * This class represents an Html label element.
+ * This class represents an HTML label element.
  * <p>
- * A label element can be connected to an {@link IDomInputNode}. If the user
- * clicks on the label, the connected input node will be selected. If the input
- * node is a text input, the input will be focused. If the input is a check-box
- * or radio-button, the <i>checked</i> state of the input will be toggled.
```
oliver-richers, line 16: The comment about checkboxes and radiobuttons was a lie. ;) I was only true for native checkboxes and radiobuttons, which we don't use anymore.

platform-dom/src/main/java/com/softicar/platform/dom/parent/IDomParentElement.java
```diff
@@ -258,16 +258,4 @@ default IDomTextNode prependText(IDisplayString text) {
 	 * Returns a collection with all child nodes.
 	 */
 	List<IDomNode> getChildren();
-
-	// -------------------------------- RECURSIVE TREE METHODS -------------------------------- //
-
-	/**
-	 * Disables all input elements contained in this element.
-	 */
-	void disableAllChildren();
```
oliver-richers, line 267: These methods were never used, so why keep it?

platform-emf/src/main/java/com/softicar/platform/emf/attribute/field/day/EmfDayInput.java
```diff
@@ -37,13 +36,6 @@ public void setValue(Day value) {
 		dayInput.setValue(value);
 	}
 
-	@Override
-	public IDomElement setEnabled(boolean enabled) {
```
oliver-richers, line 41: The method IEmfInput.setEnabled(boolean) is never used, so we don't need to implement it anymore, since IEmfInput no longer extends IDomEnableable or IDomDisableable.

platform-dom/src/main/java/com/softicar/platform/dom/input/IDomDisableable.java
```diff
@@ -7,7 +7,7 @@
  *
  * @author Oliver Richers
  */
-public interface IDomEnableable extends IDomElement {
+public interface IDomDisableable extends IDomElement {
```
oliver-richers, line 10: Renamed this interface. The name is still awkward, but at least it better describes what this is all about, i.e. DOM elements that can be **disabled**.

alexander-schmidt (dismissed):

alexander-schmidt (approved):

PR #232, opened 2022-04-26
PLAT-811 Remove redundant shadow below page header

alexander-schmidt:
image
------------------------
image


oliver-richers (approved):
nice catch

PR #236, opened 2022-04-28
PLAT-815 Limit Width of Header Cells

daniel-klose:


platform-emf/src/main/resources/com/softicar/platform/emf/emf-data-table-style.css
```diff
@@ -1,6 +1,8 @@
 
 .EmfDataTableHeaderCellDiv {
 	gap: 2px;
+	max-width: 350px;
```
daniel-klose, line 4: this is arbitrary, dont know if its good or not ¯\_(ツ)_/¯

oliver-richers (changes requested):

oliver-richers (approved):
hmm, strange that this works, but the screenshot looks fine

oliver-richers (approved):
ok, after reading https://developer.mozilla.org/en-US/docs/Web/CSS/white-space, this seems to be reasonable, although I think we should rather use `pre-line` instead of `break-spaces`

platform-emf/src/main/resources/com/softicar/platform/emf/emf-data-table-style.css
```diff
@@ -1,6 +1,7 @@
 
 .EmfDataTableHeaderCellDiv {
 	gap: 2px;
+	white-space: break-spaces;
```
oliver-richers, line 4: shouldn't this rather be `pre-line` instead of `break-spaces`?

PR #240, opened 2022-04-29
PLAT-828-1 CSV Import: Ensure That Valid Business Keys Are Set

thees-k:
If `setBusinessKey()` of `EmfTableConfiguration` is called, the set business key will now be checked: Is it a unique key? Does it not contain an autoincrement column? - There are two new JUnit tests for that.

Method `validateBusinessKey()` of `EmfTableValidator` was removed, but Gradle task "check" will still fail if a non valid business key was set. - You can try it out:

Insert the following column:
```
public static final IDbKey<AGDemoInvoice> KEY_ID_NUMBER = BUILDER.addIndexKey("idIncoiceNumber", ID, INVOICE_NUMBER);
```
.. into `AGDemoInvoiceGenerated ` just above its column...
```
public static final AGDemoInvoiceTable TABLE = new AGDemoInvoiceTable(BUILDER);
```
Then in method `customizeEmfTableConfiguration()` of `AGDemoInvoiceTable` call:
```
configuration.setBusinessKey(AGDemoInvoice.KEY_ID_NUMBER);
```
Then execute the Gradle task ":platform-demo-module:test" (a sub task of task "check").


platform-emf/src/main/java/com/softicar/platform/emf/table/configuration/EmfTableConfiguration.java
```diff
@@ -503,9 +505,27 @@ private void customizeAttributesAndOrder(EmfAttributeList<R, P> attributes) {
...
 
+		checkBusinessKey(businessKey);
 		this.businessKey = businessKey;
 	}
 
+	private void checkBusinessKey(IDbKey<R> businessKey) {
+
+		if (!businessKey.isUniqueKey()) {
+			throw new SofticarDeveloperException(
+				"The set business key (%s) of table %s must be a unique key.",
```
oliver-richers, line 516:
```suggestion
				"The business key (%s) of table %s must be a unique key.",
```
IMO, the word `set` is unnecessary and might confuse the reader

platform-emf/src/main/java/com/softicar/platform/emf/table/configuration/EmfTableConfiguration.java
```diff
@@ -503,9 +505,27 @@ private void customizeAttributesAndOrder(EmfAttributeList<R, P> attributes) {
...
+				Imploder.implode(businessKey.getFields(), " & "),
+				table);
+		}
+
+		if (table.getPrimaryKey().isGenerated() && !Collections.disjoint(businessKey.getFields(), table.getPrimaryKey().getFields())) {
+			throw new SofticarDeveloperException(
+				"The set business key (%s) of table %s must not contain an autoincrement column.",
```
oliver-richers, line 523:
```suggestion
				"The business key (%s) of table %s must not contain a column of the generated primary key.",
```
we should not imply that we have a autoincrement column, that might change in the future

oliver-richers (changes requested):

oliver-richers (approved):

PR #242, opened 2022-05-02
PLAT-765-1 Migrated Some AJAX Javascript Engine Code to Typescript

oliver-richers:
- migrated ActionQueue.js
- migrated KeepAlive.js
- removed unused Fading.js
- removed unused Mutex.js
- removed unused method enableInput in DOMContext.js

Test Plan:
- startet test servlet, opening pages, popups, etc.
- grep for the following texts to ensure proper renaming:
  - 'AQ_', 'KeepAlive', 'Mutex', 'Fade', 'Fading', 'enableInput'
  - find src/ -type f | egrep 'js$|ts$|java$' | xargs grep -in AQ_


platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/ForwardDeclarations.ts
```diff
@@ -0,0 +1,7 @@
+
+declare let SESSION_TIMED_OUT: boolean;
+
+declare function SR_sendAjaxRequest(message: any, form: any): void;
```
alexander-schmidt, line 4: That `SR_` seems cryptic. When we're already at it: Can we either remove the prefix, or spell it out?

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/KeepAlive.ts
```diff
@@ -1,52 +1,57 @@
...
-	};
+			KEEP_ALIVE.schedule();
+		}
+	}
+}
 
-	var m_delay = delay;
-	var m_handler = null;
-};
+let KEEP_ALIVE = new KeepAlive(3*60*1000);
```
alexander-schmidt, line 57: We should probably extract a constant here.

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/Actions.ts
```diff
@@ -0,0 +1,42 @@
+
+interface Action {
+	execute(): void;
+}
+
+// This action represents a server request. When executed, this
```
alexander-schmidt, line 6: In most of the other source files, we mimic...
```
/**
 * Javadoc syntax
 */
```
We should probably do the same in TS-transcribed code.
Unless this was a deliberate change of style.

alexander-schmidt (changes requested):

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/ForwardDeclarations.ts
oliver-richers: not in this PR, as you can see, it is only a forward declaration of a "module" I did not want to touch, yet.

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/KeepAlive.ts
oliver-richers: I can do that.

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/Actions.ts
oliver-richers: I have to check what is the recommended type of comments in TypeScript.

oliver-richers: converted all comments to TypeDoc

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/KeepAlive.ts
oliver-richers: did so

alexander-schmidt (approved):

PR #247, opened 2022-05-06
PLAT-830-1 Migrated JavaScript Code to TypeScript

oliver-richers:
- migrated caret related code (tested by AjaxDomInputTest)
- migrated upload code (extended AjaxUploadTest)
- migrated timeout event code (implemented AjaxDomTimeoutEventTest)
- migrated drag & drop code (tested manually)
- migrated session timeout code
- migrated working indicator code
- migrated AJAX request lock code
- removed unused form reset and submit functionality
- removed useless JavaScript calls in AjaxDocumentPagePrinter
- moved Utils.js and DomElementBuilder.js into auto-complete subfolder


platform-ajax/src/main/java/com/softicar/platform/ajax/document/creation/AjaxDocumentPagePrinter.java
```diff
@@ -118,8 +118,6 @@ private void printJavascriptUpdateCode() {
 		println("<script>");
 		println("GLOBAL_init();");
 		updateCodeWriter.apply();
-		println("lock(LOCK_REASON_DOM_EVENT);");
-		println("ACTION_QUEUE.executeNextAction();");
```
oliver-richers, line 122: I tested this, the call to `ACTION_QUEUE.executeNextAction()` does nothing since the queue is always empty on document creation. This is of course to be expected. I don't know why this was called here. Its removal also makes the `lock` call unnecessary, or rather impossible.

platform-ajax/src/main/java/com/softicar/platform/ajax/engine/AjaxDomEngine.java
```diff
@@ -300,20 +300,14 @@ public void unsetClickTargetForEventDelegation(IDomNode sourceNode, DomEventType
 	@Override
 	public void scheduleTimeout(IDomTimeoutNode timeoutNode, Double seconds) {
 
-		JS_call("c.scheduleTimeout", timeoutNode.getNodeId(), Math.round(seconds * 1000));
-	}
-
-	@Override
-	public void clearLastScheduledTimeout() {
```
oliver-richers, line 307: I removed `clearLastScheduledTimeout` because it was never called, and it also does not work deterministically. If needed it should be reimplemented.

platform-ajax/src/main/java/com/softicar/platform/ajax/engine/AjaxDomEngine.java
```diff
@@ -427,22 +421,21 @@ public void executeScriptCode(String scriptCode) {
...
-	public void submitForm(IDomNode form) {
+	public void triggerUploadOnClick(IDomNode form, IDomNode triggerNode) {
 
-		JS_call("c.submitForm", form.getNodeId());
+		triggerUploadOnEvent(form, triggerNode, "onclick");
 	}
 
-	@Override
-	public void submitFormOnChange(IDomNode form, IDomNode triggerNode) {
+	private void triggerUploadOnEvent(IDomNode form, IDomNode triggerNode, String event) {
```
oliver-richers, line 435: - I removed the method `resetForm`  since I see no point providing it.
- I removed `submitForm` because it was not used, is not necessary, and does not really do what is advertised. It actually executed an AJAX upload request.
- I renamed `submitFormOnChange` to ` triggerUploadOnEvent` since it the original name was a misnomer.

alexander-schmidt (changes requested):

platform-dom/src/main/java/com/softicar/platform/dom/event/timeout/IDomTimeoutNode.java
```diff
@@ -3,12 +3,25 @@
...
+	/**
+	 * Schedules a timeout event on this {@link IDomNode}.
+	 * <p>
+	 * After the specified amount of seconds have passed, the method
+	 * {@link IDomTimeoutNode#handleTimeout()} will be called.
+	 *
+	 * @param seconds
+	 *            the seconds to wait before the timeout
+	 */
+	default void scheduleTimeout(double seconds) {
```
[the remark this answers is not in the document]
oliver-richers: no, it is actually milliseconds on JavaScript side, so, all fine IMO

alexander-schmidt (approved):

PR #250, opened 2022-05-10
PLAT-808-2 Create Session Overview Page

daniel-klose:


platform-core-module/src/main/java/com/softicar/platform/core/module/maintenance/SessionOverviewPage.java
```diff
@@ -0,0 +1,46 @@
...
+	@Override
+	public IDisplayString getTitle(SystemModuleInstance moduleInstance) {
+
+		return CoreI18n.SESSION_OVERVIEW;
+	}
+
+	@Override
+	public IEmfRole<SystemModuleInstance> getAuthorizedRole() {
+
+		return CoreRoles.MAINTENANCE_ADMINISTRATOR;
```
oliver-richers, line 32: why can you see the page, when nobody is assigned to it? we should assign the role to the admin user in the core test fixture, IMO

oliver-richers (changes requested):

platform-core-module/src/main/java/com/softicar/platform/core/module/maintenance/SessionOverviewPage.java
daniel-klose: well, I assigned myself the role under role assignment

oliver-richers (approved):

PR #251, opened 2022-05-11
PLAT-830-2 Migrated More JavaScript Code to TypeScript

oliver-richers:
- migrated value node map (added AjaxRequestInputValuesTest)
- migrated change event handling code
- migrated keyboard event handling code
- migrated context code
- migrated event delegation code
- migrated event sending code
- removed the global object
- changed all immutable global object references to `const`


platform-ajax/src/main/java/com/softicar/platform/ajax/engine/AjaxDomEngine.java
```diff
@@ -191,9 +192,7 @@ public void replaceChild(IDomNode parent, IDomNode newChild, IDomNode oldChild)
 	public void setNodeAttribute(IDomNode node, IDomAttribute attribute) {
 
 		if (attribute.getName().equals("value")) {
-			JS_call("GLOBAL.setValue", node, attribute);
-		} else if (attribute.getName().equals("checked")) {
```
oliver-richers, line 195: we have no support for native check-boxes, so this code was removed

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/AjaxEngine.ts
```diff
@@ -0,0 +1,166 @@
+// class definition of DOMContext.
```
alexander-schmidt, line 1: What is this?

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/AjaxEngine.ts
```diff
@@ -0,0 +1,166 @@
...
+class AjaxEngine {
+	private readonly nodes = new Map<number, Node>();
+	private zIndex = 100;
+
+	public allocateZIndex() {
```
alexander-schmidt, line 6: In general, we could explicitly define return types for functions. Should we?
That way, we could avoid unintended (and probably unnoticed) changes to method return types (which are currently inferred).

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/AjaxEngine.ts
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	public setMaximumZIndex(node: HTMLElement) {
+		if(node.style.zIndex != '' + this.zIndex) {
+			node.style.zIndex = this.allocateZIndex();
+		}
+	}
+
+	public getNode(nodeId: number) {
+		return this.nodes.get(nodeId) as Node;
```
alexander-schmidt, line 18: `as Node` ... is this cast still necessary if we declare the function like
`public getNode(nodeId: number): Node {`
?

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/AjaxEngine.ts
```diff
@@ -0,0 +1,166 @@
+// class definition of DOMContext.
+class AjaxEngine {
+	private readonly nodes = new Map<number, Node>();
+	private zIndex = 100;
+
+	public allocateZIndex() {
+		this.zIndex += 1;
+		return '' + this.zIndex;
```
alexander-schmidt, line 8: Why would we force a String here? Seems odd.

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/AjaxEngine.ts
```diff
@@ -0,0 +1,166 @@
...
+
+	public getNode(nodeId: number) {
+		return this.nodes.get(nodeId) as Node;
+	}
+
+	public getElement(nodeId: number) {
+		return this.nodes.get(nodeId) as HTMLElement;
+	}
+
+	// initializes the head object of this document
```
alexander-schmidt, line 25: I think we should use a homogenous comment style across all `.ts` files; and hence modify this. Same below.

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/AjaxEngine.ts
```diff
@@ -0,0 +1,166 @@
...
+		node.id = 'n' + nodeId;
+		this.nodes.set(nodeId, node);
+	}
+}
+
+const _DOM_CONTEXT_ = new AjaxEngine();
+
+// -------------------- abbreviations -------------------- //
+
+const c = _DOM_CONTEXT_;
```
alexander-schmidt, line 116: Do we somehow rely on `c` being publicly available? If we don't, I suggest to remove this... alias.

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/AjaxEngine.ts
```diff
@@ -0,0 +1,166 @@
...
+		}
+	}
+
+	private initializeNode(nodeId: number, node: HTMLElement) {
+		node.id = 'n' + nodeId;
+		this.nodes.set(nodeId, node);
+	}
+}
+
+const _DOM_CONTEXT_ = new AjaxEngine();
```
alexander-schmidt, line 112: ...if it's in fact the engine, we should probably just rename the constant to `_ENGINE_` or `_AJAX_ENGINE_`. Because what's the point of having a synonym?

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/Utils.ts
```diff
@@ -0,0 +1,6 @@
+/**
```
alexander-schmidt, line 1: I kind of dislike the implied pattern of collecting all sorts of utility functions in this file. That won't scale well.
Granted, we could just refactor this in the future.

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/ValueNodeMap.ts
```diff
@@ -0,0 +1,114 @@
...
+
+	public isValueChanged(node: HTMLElement) {
+		return this.getState(node).isValueChanged();
+	}
+
+	public setValue(node: HTMLInputElement, value: string) {
+		node.value = value;
+		this.getState(node).assumeValue(value);
+
+		// FIXME this ad-hoc call looks like a hack
```
alexander-schmidt, line 36: Do we already have a story to re-implement ACs? If so, we could reference it here.

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/ValueNodeMap.ts
```diff
@@ -0,0 +1,114 @@
...
+	public setSelectedOptions(select: HTMLSelectElement, options: HTMLOptionElement[]) {
+		let type = select.type;
+		if(type == "select-one") {
+			options[0].selected = true;
+		} else if(type == "select-multiple") {
+			for(let option of select.options) {
+				option.selected = options.includes(option);
+			}
+		} else {
+			alert("Unknown select type " + type + ".");
```
alexander-schmidt, line 49: - Punctuation seems odd... should be `"Unknown select type: " + type` or so.
- We should probably add the ID of the offending node to the error message.
- Why would we use a native alert in the first place? (rather than logging to console, or using `throw new Error("...")` much like a couple of lines below)

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/ValueNodeMap.ts
```diff
@@ -0,0 +1,114 @@
...
+
+		this.getState(select).assumeValue(this.getState(select).getCurrentValue());
+	}
+
+	private getState(node: HTMLElement) {
+		let state = this.states.get(node);
+		if(state) {
+			return state;
+		} else {
+			throw new Error("Internal error: Unexpected value node.");
```
alexander-schmidt, line 60: `Unexpected value node` does not seem to describe the actual problem. I would have expected sth like "...unknown state...".

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/event/EventToClickDelegation.ts
```diff
@@ -0,0 +1,44 @@
+const EVENT_DELEGATIONS = new Map<HTMLElement, EventToClickDelegator>();
+
+function getOrCreateEventDelegator(element: HTMLElement) {
```
alexander-schmidt, line 3: We seem to randomly mix OOP and global functions. I think we should improve upon that in the future.

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/event/ChangeEvents.ts
```diff
@@ -0,0 +1,29 @@
...
+		}
+
+		// only send or delegate event if value really changed
+		if(VALUE_NODE_MAP.isValueChanged(node)) {
+			sendOrDelegateEvent(node, event, event.type);
+		}
+	}
+}
+
+const CHANGE_EVENT_MANAGER = new ChangeEventManager();
```
alexander-schmidt, line 29: Should we somehow advertize this as the singleton handle to this class? We could add a comment here, or above the class.
Applies to other classes as well.

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/event/KeyboardEvents.ts
```diff
@@ -0,0 +1,144 @@
...
+			case 'SPACE':  return DOM_VK_SPACE;
+			case 'TAB':    return DOM_VK_TAB;
+		}
+		throw new Error("Internal error: Unsupported keyboard event name.");
+	}
+}
+
+const KEYBOARD_EVENT_MANAGER = new KeyboardEventManager();
+
+class KeyboardEventHandler {
```
alexander-schmidt, line 47: Apparently, there's a way to emulate private classes in TS:
https://stackoverflow.com/a/17163506
...not exactly necessary here - but interesting nevertheless.

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/event/KeyboardEvents.ts
```diff
@@ -0,0 +1,144 @@
...
+	public constructor(node: HTMLElement) {
+		this.node = node;
+	}
+	
+	public install() {
+		if(!this.node.onkeydown && !this.node.onkeyup) {
+			this.node.onkeydown = event => this.handleKeyDown(event);
+			this.node.onkeyup = event => this.handleKeyUp(event);
+		} else {
+			console.log('warning: skipped installation of keyboard event listeners');
```
alexander-schmidt, line 64:
```suggestion
			console.log('Warning: Skipped installation of keyboard event listeners.');
```

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/event/KeyboardEvents.ts
```diff
@@ -0,0 +1,144 @@
...
+	}
+}
+
+class CssClassApplier {
+	private readonly node: HTMLElement;
+	private readonly classes: string[];
+
+	public constructor(node: HTMLElement, classes: string) {
+		this.node = node;
+		this.classes = classes != ""? classes.split(" ") : [];
```
alexander-schmidt, line 134:
```suggestion
		this.classes = classes != "" ? classes.split(" ") : [];
```
...equal spacing around both "?" and ":".

platform-dom/src/main/java/com/softicar/platform/dom/input/DomSelect.java
```diff
@@ -62,6 +62,16 @@ public final O getSelectedOption() {
 		}
 	}
 
+	/**
+	 * Returns a list of all selected option objects.
```
alexander-schmidt, line 66:
```suggestion
	 * Returns all selected option objects.
```
...it's not quite a "list".

alexander-schmidt (changes requested):

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/AjaxEngine.ts
oliver-richers: oups, removed

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/AjaxEngine.ts
oliver-richers: I don't think that would improve the code, but rather increase unnecessary clutter....if you change the implicit return type and that would make calling code invalid, we will notice it.

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/AjaxEngine.ts
oliver-richers: yes, because, `get` returns `Node | null`, so we have to cast

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/AjaxEngine.ts
oliver-richers: because the result is assigned to `style.zIndex`, otherwise it would be an error on the caller site

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/AjaxEngine.ts
oliver-richers: did so

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/AjaxEngine.ts
oliver-richers: yes, I believe the DomAjaxEngine uses it but I will rework this in a follow-up PR.

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/AjaxEngine.ts
oliver-richers: the point is that I am waiting for your PR to get ready :P `_DOM_CONTEXT_` is used in `DomPopupEngine.js`

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/Utils.ts
oliver-richers: then please give a suggestion, I will probably put all methods called by `AjaxDomEngine.java` into `AjaxEngine.ts`, so this is probably a non-issue

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/ValueNodeMap.ts
oliver-richers: added reference to PLAT-853

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/ValueNodeMap.ts
oliver-richers: improved

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/event/EventToClickDelegation.ts
oliver-richers: planned for next PR

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/event/ChangeEvents.ts
oliver-richers: my plan is to gather all global variables in the instance of `AjaxEngine.ts`.

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/event/KeyboardEvents.ts
oliver-richers: yes, indeed, but I think modularization is for another story

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/event/KeyboardEvents.ts
oliver-richers: done

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/event/KeyboardEvents.ts
oliver-richers: in Java we have no space before the `?` so I did the same here

platform-dom/src/main/java/com/softicar/platform/dom/input/DomSelect.java
oliver-richers: did so

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/AjaxEngine.ts
alexander-schmidt: Hm. Fair point.

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/AjaxEngine.ts
alexander-schmidt: k

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/AjaxEngine.ts
alexander-schmidt: i see

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/event/ChangeEvents.ts
alexander-schmidt: Sounds reasonable

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/event/KeyboardEvents.ts
alexander-schmidt: Indeed, I never realized this. Never mind.

PR #256, opened 2022-05-16
PLAT-855 Adapt SessionTimeoutDialogMessage

daniel-klose:
image


platform-common/src/main/java/com/softicar/platform/common/core/CommonCoreI18n.java
```diff
@@ -15,6 +15,8 @@ public interface CommonCoreI18n {
 		.de("Zifferngruppe zu kurz.");
 	I18n1 ILLEGAL_CHARACTERS_FOR_DECIMAL_NUMBER_ARG1 = new I18n1("Illegal characters for decimal number: %s")//
 		.de("Unzulässige Zeichen für Dezimalzahl: %s");
+	I18n0 RETURN_TO_LOGIN = new I18n0("Return to login")//
```
alexander-schmidt, line 18:
```suggestion
	I18n0 RETURN_TO_LOGIN = new I18n0("Return to Login")//
```

alexander-schmidt (approved):

PR #260, opened 2022-05-16
PLAT-857 Correctly Style Footer Rows in DomDataTable

daniel-klose:
- Removed hardstyled text-alignment
  - Alignment is now default, which is center
  - Added convenience method for NumericCells in footer, which has right-alignment
  - Example:  
image


oliver-richers (approved):
nice

PR #261, opened 2022-05-17
PLAT-829-2 Finally Solved The Dreaded Double-Event Problem

oliver-richers:
- implement redundancy check for AjaxRequestMessage
- removed now useless AjaxRequestLock
- removed now useless AjaxRequestManager

Test Plan:
- see AjaxMultiEventTest


platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/request/AjaxRequestMessage.ts
```diff
@@ -90,7 +90,55 @@ class AjaxRequestMessage {
 		return new AjaxRequestMessageEncoder(this.data).encodeToHex();
 	}
 
-	// ------------------------------ private ------------------------------ //
+	// ------------------------------ redundancy ------------------------------ //
+
+	public isRedundantTo(other: AjaxRequestMessage) {
```
oliver-richers, line 95: this method defines which requests will be dropped if other requests are queued or sent

alexander-schmidt (changes requested):

alexander-schmidt (approved):

PR #262, opened 2022-05-17
PLAT-808-5 Implement Maintenance Logic

daniel-klose:


platform-core-module/src/main/java/com/softicar/platform/core/module/page/service/login/PageServiceLoginDiv.java
```diff
@@ -48,13 +49,15 @@ private class LoginForm extends DomForm {
 		private final LoginButton loginButton;
 		private final UserInput userInput;
 		private final PasswordInput passwordInput;
+		private final MaintenanceElement maintenanceDiv;
```
oliver-richers, line 52: IMO, should be `MaintenanceElement maintenanceElement;` or `MaintenanceDiv maintenanceDiv;`

oliver-richers (approved):
nice

platform-core-module/src/main/java/com/softicar/platform/core/module/page/service/login/PageServiceLoginDiv.java
daniel-klose: can do in another cleanup revision

PR #263, opened 2022-05-17
PLAT-826-2 Limit drag area for popups

alexander-schmidt:
- It is now possible to limit/constrain the area in which a node can be dragged.
  - `DomPopup` elements can no longer be dragged to be completely outside of their respective `IDomPopupMaximizationContext`.
- `IDomEngine.makeDraggable`:
  - Introduced nullable `limitingNode` parameter.
  - Renamed parameters to be less confusing. The same renaming was performed in `DragAndDrop.ts`.
  - Added Javadoc.
- Removed `IDomElement.makeDraggable*` methods.

Peek 2022-05-17 19-14


platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/DragAndDrop.ts
```diff
@@ -1,46 +1,50 @@
 
-function makeDraggable(draggedNode: HTMLElement, initNode: HTMLElement, notifyOnDrop: boolean) {
-	new DragContext(draggedNode, notifyOnDrop).setup(initNode);
+function makeDraggable(movedNode: HTMLElement, draggedNode: HTMLElement, limitingNode: HTMLElement | null, notifyOnDrop: boolean) {
```
oliver-richers, line 2: I find the renaming of the `draggedNode` and `initNode` confusing, maybe `draggedNode` and `dragInitiationNode`?

platform-ajax/src/test/java/com/softicar/platform/ajax/drag/and/drop/AjaxDragAndDropTest.java
```diff
@@ -43,7 +43,7 @@ public TestDiv() {
 			this.dragTargetDiv = appendChild(new Div(200, 100));
 			this.dragAnchorDiv = dragTargetDiv.appendChild(new Div(50, 50));
 
-			getDomEngine().makeDraggable(dragTargetDiv, dragAnchorDiv);
+			getDomEngine().makeDraggable(dragTargetDiv, dragAnchorDiv, null);
```
oliver-richers, line 46: it seems that this `null` value is not properly converted to a `null` value in JavaScript, we should try to fix this code in `DomAjaxEngine`. At least `AjaxDragAndDropTest` passes when change to plain `"null"`.
image

Okay, I changed this locally and all check pass.

platform-ajax/src/main/java/com/softicar/platform/ajax/engine/AjaxDomEngine.java
```diff
@@ -310,10 +310,10 @@ public void scheduleTimeout(IDomTimeoutNode timeoutNode, Double seconds) {
 	}
 
 	@Override
-	public void makeDraggable(IDomNode draggedNode, IDomNode initNode) {
+	public void makeDraggable(IDomNode movedNode, IDomNode draggedNode, IDomNode limitingNode) {
```
oliver-richers, line 313: I find the renaming of the `draggedNode` and `initNode` confusing, maybe `draggedNode` and `dragInitiationNode`?

oliver-richers (changes requested):

platform-ajax/src/test/java/com/softicar/platform/ajax/drag/and/drop/AjaxDragAndDropTest.java
alexander-schmidt: Seems to work, thanks for the proposal

platform-ajax/src/main/typescript/com/softicar/platform/ajax/engine/DragAndDrop.ts
alexander-schmidt: To me, the "initiation" term seems quite puzzling. It does not seem to imply that the user interacts with this thing (which _should_ be implied, IMO... otherwise, we need a lengthy Javadoc explanation).
That being said, I think that calling the first node "dragged node" is a misnomer: The user does not (directly) perform a drag-and-drop operation on this node. It is merely _moved_ because they perform a drag-and-drop operation on a _different_ node.
Please reconsider.

alexander-schmidt: Renamed stuff as discussed.

platform-ajax/src/main/java/com/softicar/platform/ajax/engine/AjaxDomEngine.java
alexander-schmidt: Renamed stuff as discussed.

oliver-richers (approved):

PR #268, opened 2022-05-19
PLAT-852 Fixed EmfForm width

daniel-klose:
image


oliver-richers (approved):
not a big fan of `width: 100%;` but if it solves the problem, i am fine with it :)

PR #269, opened 2022-05-19
PLAT-854 Fix Workflow Manager not properly repainting

daniel-klose:


platform-workflow-module/src/main/java/com/softicar/platform/workflow/module/workflow/management/display/element/WorkflowNodeDisplayElement.java
```diff
@@ -43,7 +43,7 @@ public void handleDrop(IDomDropEvent dropParameters) {
 
 		int x = (dropParameters.getDropX() + 5) / 10 * 10;
 		int y = (dropParameters.getDropY() + 5) / 10 * 10;
-		if (x >= 0) {
+		if (x > 0 && y > 0) {
```
oliver-richers, line 46: why is 0 not allowed?

oliver-richers (approved):
acceptable, but please answer the question

platform-workflow-module/src/main/java/com/softicar/platform/workflow/module/workflow/management/display/element/WorkflowNodeDisplayElement.java
daniel-klose: its a bit close to the edges, no particular reason

oliver-richers: well from a purity standpoint I don't like it, but in practice it would not matter, so, ok

PR #277, opened 2022-05-20
PLAT-864 Set proper comparator for column drop-down

daniel-klose:
Before:
image
After:
image

### DIFF
--- a/platform-dom/src/main/java/com/softicar/platform/dom/elements/select/value/simple/display/DomSimpleValueSelectDefaultValueComparator.java
+++ b/platform-dom/src/main/java/com/softicar/platform/dom/elements/select/value/simple/display/DomSimpleValueSelectDefaultValueComparator.java
@@ -5,6 +5,7 @@ import com.softicar.platform.common.core.i18n.IDisplayable;
 import com.softicar.platform.common.core.item.BasicItemComparator;
 import com.softicar.platform.common.core.item.IBasicItem;
 import com.softicar.platform.dom.elements.select.value.simple.DomSimpleValueSelect;
+import java.math.BigDecimal;
 import java.util.Arrays;
 import java.util.Comparator;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


platform-dom/src/main/java/com/softicar/platform/dom/elements/select/value/simple/display/DomSimpleValueSelectDefaultValueComparator.java
```diff
@@ -59,22 +60,33 @@ public int compare(V first, V second) {
...
+				return compareNullsFirst(Integer::compareTo, (Integer) first, (Integer) second);
+			} else if (isArgumentsCastableTo(first, second, Float.class)) {
+				return compareNullsFirst(Float::compareTo, (Float) first, (Float) second);
+			} else if (isArgumentsCastableTo(first, second, Double.class)) {
+				return compareNullsFirst(Double::compareTo, (Double) first, (Double) second);
+			} else if (isArgumentsCastableTo(first, second, BigDecimal.class)) {
+				return compareNullsFirst(BigDecimal::compareTo, (BigDecimal) first, (BigDecimal) second);
+			} else if (isArgumentsCastableTo(first, second, Long.class)) {
+				return compareNullsFirst(Long::compareTo, (Long) first, (Long) second);
+			}
```
daniel-klose, line 87: I know this is ugly, but every implementation of `Number` has a different `compareTo`, therefore I dont really know how to do this better than simple `else if` blocks

platform-dom/src/main/java/com/softicar/platform/dom/elements/select/value/simple/display/DomSimpleValueSelectDefaultValueComparator.java
```diff
@@ -59,22 +60,33 @@ public int compare(V first, V second) {
...
-					result = Comparator.nullsFirst(comparator).compare((IBasicItem) first, (IBasicItem) second);
+					result = compareNullsFirst(comparator, (IBasicItem) first, (IBasicItem) second);
 				}
 			}
 
 			return result;
 		}
 
 		else {
+			if (isArgumentsCastableTo(first, second, Integer.class)) {
```
oliver-richers, line 77: for everything but BigDecimal, you could use Number.doubleValue
```
if (isArgumentsCastableTo(first, second, Number.class)) {
    return compareNullsFirst(Double.compareTo, toNumberDoubleValue(first), toNumberDoubleValue(second));
}
```
and event for BigDecimal, i think Number.doubleValue might suffice, if not then do the BigDecimal check before the Number check.

oliver-richers (changes requested):

platform-dom/src/main/java/com/softicar/platform/dom/elements/select/value/simple/display/DomSimpleValueSelectDefaultValueComparator.java
```diff
@@ -59,22 +61,32 @@ public int compare(V first, V second) {
...
 			return result;
 		}
 
 		else {
+			if (isArgumentsCastableTo(first, second, BigDecimal.class)) {
+				return compareNullsFirst(BigDecimal::compareTo, (BigDecimal) first, (BigDecimal) second);
+			} else if (isArgumentsCastableTo(first, second, Number.class)) {
+				Number firstNumber = (Number) first;
+				Number secondNumber = (Number) second;
+				return compareNullsFirst(
```
oliver-richers, line 83: this is not necessary, the test `isArgumentsCastableTo` would have failed, if one of the values was `null`

oliver-richers: ok, that method is strange

oliver-richers (approved):

PR #281, opened 2022-05-20
PLAT-880 Rename 'dialog popups' to 'dialogs'

alexander-schmidt:


oliver-richers (approved):
some team members will not like this change ;)

PR #282, opened 2022-05-20
PLAT-875 Append modal popups and backdrops to the closest maximization context

alexander-schmidt:
- Modal backdrops are no longer displayed full-page, by default. Instead, the closest maximization context (if any) is preferred as a parent.
- Popups spawned from modal popups are now also appended to the closest maximization context (instead of being appended to the body).

This has the following effects:
- Modal backdrops spawned from a page are displayed inside the `PageContentDiv`. They are no longer overlays above the entire viewport.
- Navigation and page header can be interacted with, while a modal popup spawned from a page is present.
- If a modal popup is spawned from outside the page (e.g. from the logout button), backdrop and popup will still be appended to the body.
- Fixed: Popups that were spawned from an EMF management table action popover are now in the same z-index sequence as non-modal popups spawned from the page.

Peek 2022-05-20 17-25


oliver-richers (approved):
nice, some unit test would have been great

oliver-richers (approved):

PR #286, opened 2022-05-23
PLAT-851 Add EmfDataTable filter for (not-) empty

daniel-klose:


platform-emf/src/main/java/com/softicar/platform/emf/data/table/filter/string/EmfDataTableStringFilterNode.java
```diff
@@ -40,14 +40,29 @@ public IEmfDataTableFilter<R> createFilter() {
...
+			}
 			return new EmfDataTableNopFilter<>(resetter);
 		}
 	}
 
+	@Override
+	public void refresh() {
+
+		super.refresh();
+		filterInput.refresh(filterTypeSelect.getSelectedValue());
```
oliver-richers, line 63: not 100% sure if this is the cleanest solution, but I have no other suggestion, just saying

oliver-richers (approved):

platform-emf/src/main/java/com/softicar/platform/emf/data/table/filter/string/EmfDataTableStringFilterNode.java
daniel-klose: definitely not the cleanest, but the best without bigger refactorings imho

oliver-richers (approved):

PR #290, opened 2022-05-24
PLAT-835 Implemented System Events

oliver-richers:
image

image


platform-dom/src/main/java/com/softicar/platform/dom/json/DomJsonBooleanDisplay.java
```diff
@@ -0,0 +1,14 @@
+package com.softicar.platform.dom.json;
+
+import com.google.gson.JsonPrimitive;
+import com.softicar.platform.dom.DomCssClasses;
+import com.softicar.platform.dom.elements.DomDiv;
+
+class DomJsonBooleanDisplay extends DomDiv {
+
+	public DomJsonBooleanDisplay(JsonPrimitive primtive) {
```
daniel-klose, line 9: primtive -> primitive

platform-core-module/src/main/java/com/softicar/platform/core/module/event/SystemEventPage.java
```diff
@@ -0,0 +1,39 @@
...
+	@Override
+	public Class<CoreModule> getModuleClass() {
+
+		return CoreModule.class;
+	}
+
+	@Override
+	public EmfPagePath getPagePath(EmfPagePath modulePath) {
+
+		return modulePath.append(CoreI18n.MAINTENANCE);
```
daniel-klose, line 25: We should definitely rename `MAINTENANCE` to something more sensible.
Currently, theres a maintenance page, session overview page and after this PR the System event page.
It should definitely not stay with the name `MAINTENANCE`

daniel-klose (changes requested):

platform-core-module/src/main/java/com/softicar/platform/core/module/event/SystemEventPage.java
oliver-richers: suggestions? I think the name is ok, tbh.

platform-core-module/src/main/java/com/softicar/platform/core/module/event/SystemEventConfirmAction.java
```diff
@@ -0,0 +1,61 @@
...
+import com.softicar.platform.dom.elements.DomElementsImages;
+import com.softicar.platform.emf.action.AbstractEmfButtonAction;
+import com.softicar.platform.emf.action.IEmfManagementAction;
+import com.softicar.platform.emf.authorization.role.IEmfRole;
+import com.softicar.platform.emf.form.IEmfFormBody;
+import com.softicar.platform.emf.predicate.IEmfPredicate;
+import java.util.Optional;
+import java.util.function.Supplier;
+
+public class SystemEventConfirmAction extends AbstractEmfButtonAction<AGSystemEvent> implements IEmfManagementAction<AGSystemEvent> {
```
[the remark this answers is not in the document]
oliver-richers: Well this heavily depends on the assumed workflow. My idea is that the system administrator just confirms that they have seen the problem, and took action, will take actions, of just ignore the notification. Don't forget this can be an error message or a warning.

daniel-klose (dismissed):

platform-core-module/src/main/java/com/softicar/platform/core/module/event/AGSystemEvent.sqml
```diff
@@ -0,0 +1,14 @@
+PACKAGE com.softicar.platform.core.module.event
+
+IMPORT com.softicar.platform.core.module.event.severity.AGSystemEventSeverity
+IMPORT com.softicar.platform.core.module.user.AGUser
+
+TABLE AGSystemEvent "Core.SystemEvent" {
+	PK Integer id = SERIAL
+	Boolean needsAttention
```
alexander-schmidt, line 8: I think the consistency in our terminology could be improved:
The button labelled `Confirm All` flips the `Needs Attention` flag. That's not intuitive - we're using two different terms for the same concept.
I therefore suggest to rename this column to `Needs Confirmation`.

platform-core-module/src/main/java/com/softicar/platform/core/module/event/AGSystemEvent.sqml
```diff
@@ -0,0 +1,14 @@
...
+	AGSystemEventSeverity severity
+	AGUser causedBy
+	DayTime causedAt
```
alexander-schmidt, line 11: I think we should swap `causedAt` and `causedBy`.
The sequence in which things happened (as denoted via `causedAt`) is usually more important than finding out who caused them (as denoted via `causedBy`). Or, the other way around: In absence of temporal ordering of events, the "who" would be almost irrelevant.
The column order in `AGTransaction` might have emerged from the same thought.

platform-core-module/src/main/java/com/softicar/platform/core/module/event/SystemEventPage.java
alexander-schmidt: I think we should ask the Ops dudes about this.

alexander-schmidt (changes requested):

platform-core-module/src/main/java/com/softicar/platform/core/module/event/AGSystemEvent.sqml
oliver-richers: as you desire :)

platform-core-module/src/main/java/com/softicar/platform/core/module/event/SystemEventPage.java
oliver-richers: did so, see screenshots

alexander-schmidt (approved):

PR #302, opened 2022-06-08
PLAT-735-2 All IDomValueInputs Now Support Change Callback

oliver-richers:
- removed unused features from `DomTimeInput`
- added new `NullaryVoidFunctionList`
- converted `AbstractEmfInputDiv` to more generic `AbstractDomValueInput`
- simplified `DomDayInput`, `DomDayTimeInput` and `DomTimeInput`


daniel-klose (dismissed):

platform-dom/src/main/java/com/softicar/platform/dom/input/AbstractDomTextualInput.java
```diff
@@ -0,0 +1,108 @@
...
+	@Override
+	public IDomTextualInput setPlaceholder(IDisplayString placeholder) {
+
+		setAttribute("placeholder", placeholder.toString());
+		return this;
+	}
+
+	// -------------------------------- read-only -------------------------------- //
+
+	/**
```
alexander-schmidt, line 74: There's idential Javadoc in the implemented interface. I think we should remove it here.

platform-dom/src/main/java/com/softicar/platform/dom/input/AbstractDomTextualInput.java
```diff
@@ -0,0 +1,108 @@
...
+	 * Makes this {@link IDomTextualInput} read-only.
+	 */
+	@Override
+	public void setReadonly(boolean readonly) {
+
+		setAttribute("readonly", readonly? "" : null);
+	}
+
+	/**
```
alexander-schmidt, line 83: There's idential Javadoc in the implemented interface. I think we should remove it here.

platform-emf/src/main/java/com/softicar/platform/emf/action/EmfPromptActionInput.java
```diff
@@ -92,7 +91,7 @@ public void appendRow(IDisplayString title, IDomNode node) {
 	@Override
 	public DomTextArea appendTextArea(IDisplayString title) {
 
-		EmfMultilineStringInput textArea = new EmfMultilineStringInput();
+		DomTextArea textArea = new DomTextArea().setRowCount(5);
```
alexander-schmidt, line 94: I think we should extract that `5` to a descriptively-named constant.

platform-emf/src/main/java/com/softicar/platform/emf/attribute/field/bool/EmfBooleanInput.java
```diff
@@ -2,15 +2,14 @@
...
 import com.softicar.platform.dom.elements.checkbox.DomCheckbox;
 import com.softicar.platform.emf.attribute.input.IEmfInput;
-import java.util.Objects;
 import java.util.Optional;
-import java.util.function.Consumer;
 
 public class EmfBooleanInput extends DomCheckbox implements IEmfInput<Boolean> {
 
-	private INullaryVoidFunction callback = INullaryVoidFunction.NO_OPERATION;
+	private final NullaryVoidFunctionList callbacks = new NullaryVoidFunctionList();
```
alexander-schmidt, line 12: Hmm, we usually initialize fields in the CTOR.

platform-emf/src/main/java/com/softicar/platform/emf/data/table/export/conversion/TableExportTypedNodeConverter.java
```diff
@@ -285,7 +285,12 @@ else if (node instanceof ITableExportTypedNode) {
 
 		else if (node instanceof DomTextArea) {
 			debugPrint(currentDepth, "DomTextArea");
-			values.add(createStringExportNodeValue(((DomTextArea) node).getInputTextTrimmed()));
+			values.add(createStringExportNodeValue(((DomTextArea) node).getValueTextTrimmed()));
+		}
+
+		else if (node instanceof DomTextInput) {
```
alexander-schmidt, line 291: I don't think this should reside in the `PARENT ELEMENTS` block. Should be moved back, down to the `NON-PARENT ELEMENTS` block.

platform-emf/src/main/java/com/softicar/platform/emf/entity/table/column/handler/EmfDisplayStringToTextAreaColumnHandler.java
```diff
@@ -12,9 +12,9 @@ public class EmfDisplayStringToTextAreaColumnHandler extends EmfDataTableValueBa
 	@Override
 	public void buildCell(IEmfDataTableCell<?, IDisplayString> cell, IDisplayString value) {
 
-		DomDiv content = new DomDiv();
+		var content = new DomDiv();
 		content.addCssClass(EmfCssClasses.EMF_DISPLAY_STRING_TO_TEXT_AREA_COLUMN_HANDLER);
-		DomTextArea textArea = new DomTextArea(value.toString(), 3, 40);
+		var textArea = new DomTextArea(value.toString()).setSize(3, 40);
```
alexander-schmidt, line 17: Kind of unrelated... but I think those numbers should be constants.

platform-dom/src/main/java/com/softicar/platform/dom/elements/DomTextArea.java
```diff
@@ -102,15 +47,10 @@ public DomTextArea setColCount(int col) {
...
-		setAttribute("readonly", "readonly");
-		return this;
-	}
-
 	/**
-	 * Enables/Disables wrapping behavior of text area fields. WARNING: This is
-	 * a non-standard attribute but it works in most browsers.
+	 * Toggles the wrapping behavior of this {@link DomTextArea}.
+	 * <p>
+	 * WARNING: This is a non-standard attribute but it works in most browsers.
```
alexander-schmidt, line 53: Uhm... no. It _is_ a standard attribute - we just use non-standard values, for whatever reason.
cf.:
- https://www.w3schools.com/tags/att_textarea_wrap.asp
- https://caniuse.com/mdn-html_elements_textarea_wrap

We should probably change this to either `hard` or `soft` instead of `on`, and to unset the attribute instead of setting it to `off`.

platform-dom/src/main/java/com/softicar/platform/dom/elements/input/AbstractDomNumberInput.java
```diff
@@ -79,14 +91,18 @@ public Optional<T> getValueNoThrow() {
 	protected abstract String formatValue(T value);
 
 	/**
-	 * Parses the input text into a value.
+	 * Parses the text into a value.
 	 * <p>
 	 * If parsing fails, a exception will be thrown.
```
alexander-schmidt, line 96:
```suggestion
	 * If parsing fails, an exception will be thrown.
```

platform-dom/src/main/java/com/softicar/platform/dom/elements/testing/engine/IDomTestExecutionEngine.java
```diff
@@ -58,14 +58,15 @@ public interface IDomTestExecutionEngine extends TestRule {
 
 	/**
 	 * Inputs the specified text into the given {@link IDomTextualInput} and
-	 * triggers an {@link IDomEvent} of type {@link DomEventType#TAB}.
+	 * triggers an {@link IDomEvent} of type {@link DomEventType#TAB} and
```
alexander-schmidt, line 61: "an event" ... of two distinct types? I am confused: Does it trigger two events, or is that XOR? Should be clarified.

alexander-schmidt (changes requested):

platform-dom/src/main/java/com/softicar/platform/dom/input/IDomValueInput.java
```diff
@@ -67,4 +77,14 @@ default Optional<V> getValueNoThrow() {
 			return Optional.empty();
 		}
 	}
+
+	/**
+	 * Registers the given callback to be notified when the value changes.
+	 * <p>
+	 * All callback functions are called in the order that they where added.
```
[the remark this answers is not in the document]
oliver-richers: Hmm, not quite. I added the comment to `setValueAndHandleChangeCallback` but I also retained the text here with a better explanantion.

platform-dom/src/main/java/com/softicar/platform/dom/input/AbstractDomValueInput.java
```diff
@@ -0,0 +1,75 @@
+package com.softicar.platform.dom.input;
+
+import com.softicar.platform.common.core.interfaces.INullaryVoidFunction;
+import com.softicar.platform.common.core.interfaces.NullaryVoidFunctionList;
+import com.softicar.platform.dom.elements.DomDiv;
+
+public abstract class AbstractDomValueInput<V> extends DomDiv implements IDomValueInput<V> {
```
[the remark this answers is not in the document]
oliver-richers: strange that class does not exist anymore, but I added  Javadoc to the protected methods in `AbstractDomValueInputElement`.

platform-dom/src/main/java/com/softicar/platform/dom/input/AbstractDomTextualInput.java
oliver-richers: indeed, removed

platform-emf/src/main/java/com/softicar/platform/emf/attribute/field/bool/EmfBooleanInput.java
oliver-richers: ok

platform-emf/src/main/java/com/softicar/platform/emf/entity/table/column/handler/EmfDisplayStringToTextAreaColumnHandler.java
oliver-richers: - I see not big benefit in making them constants, these numbers have no "semantics" and are not reused.
- Actually, we should probably remove all those calls and make this over CSS.

platform-emf/src/main/java/com/softicar/platform/emf/action/EmfPromptActionInput.java
oliver-richers: - I see no benefit in making this a constant, this number has no "semantics" and is not reused.
- Actually, we should probably remove all those calls and make this over CSS.

platform-dom/src/main/java/com/softicar/platform/dom/elements/DomTextArea.java
oliver-richers: I don't know. Then the boolean flag does not make sense. Since this method was not use here an in EAS, I removed it.

platform-dom/src/main/java/com/softicar/platform/dom/elements/input/AbstractDomNumberInput.java
oliver-richers: done

platform-dom/src/main/java/com/softicar/platform/dom/elements/testing/engine/IDomTestExecutionEngine.java
oliver-richers: fixed

alexander-schmidt (approved):

PR #309, opened 2022-06-10
PLAT-924 Fix popup positioning in (screen) edge cases

alexander-schmidt:


oliver-richers (approved):
ok, as tested during session

PR #315, opened 2022-06-14
PLAT-935-1 Implemented StoredFileContentMigrationProgram

oliver-richers:


platform-core-module/src/main/java/com/softicar/platform/core/module/file/stored/content/StoredFileContentMigrationExecutor.java
```diff
@@ -0,0 +1,107 @@
...
+	public StoredFileContentMigrationExecutor() {
+
+		this.exceptionsCollector = new ExceptionsCollector();
+	}
+
+	public void execute() {
+
+		AGStoredFileServer primaryFileServer = AGCoreModuleInstance.getInstance().getPrimaryFileServer();
+		if (primaryFileServer != null && primaryFileServer.isActive()) {
+			this.primaryStore = new StoredFileSmbContentStore(primaryFileServer);
```
alexander-schmidt, line 31: For the sake of eymmetry:

I suggest to either add a `File` infix here as well (so this would become `primaryFileStore`), **or** to remove the infixin the line above (so `primaryFileServer` would become `primaryServer`).

**Alternatively**, we could create a private method `loadPrimaryStore` (just like the existing `loadSecondaryStores`) in this class.

platform-core-module/src/main/java/com/softicar/platform/core/module/file/stored/content/StoredFileContentMigrationExecutor.java
```diff
@@ -0,0 +1,107 @@
...
+
+	private void copyFile(StoredFileContentName contentName) {
+
+		for (StoredFileSmbContentStore store: secondaryStores) {
+			if (store.exists(contentName.getFullFilename())) {
+				copyFileFromStore(contentName, store);
+				return;
+			}
+		}
+		throw new RuntimeException("Failed to find %s on secondary stores.".formatted(contentName));
```
alexander-schmidt, line 80:
```suggestion
		throw new RuntimeException("Failed to find '%s' on secondary stores.".formatted(contentName));
```

alexander-schmidt (changes requested):

oliver-richers:
I seems we will not need this.

PR #326, opened 2022-06-22
PLAT-913 Fix DomDayInput display without label

daniel-klose:
image


platform-dom/src/main/java/com/softicar/platform/dom/elements/time/day/AbstractDomDayPopupButton.java
```diff
@@ -37,7 +37,11 @@ public AbstractDomDayPopupButton() {
 
 	private void updateLabel() {
 
-		setLabel(IDisplayString.create(showLabel? chooser.getDay().toISOString() : ""));
+		if (showLabel) {
```
alexander-schmidt, line 40: I think we can revert the changes to this file. Should not make a difference since the change in DomButton was introduced.

daniel-klose: this change makes this class more readable imho, without having to know that `setLabel("")` removes the label

oliver-richers (approved):

PR #331, opened 2022-06-27
PLAT-961 Allow subsampling to boost PDF preview rendering

alexander-schmidt:
For a well-known test PDF, this change reduces the rendering time per page from ~1500 ms to ~320 ms, at the cost of an apparently marginal loss in quality.


oliver-richers (approved):
nice

PR #334, opened 2022-06-29
PLAT-970 Add migration to platform v20

alexander-schmidt:


oliver-richers (dismissed):
should work, as demonstrated

daniel-klose (approved):

PR #336, opened 2022-06-29
PLAT-980 EmfPasswordInput: Use a regular button as 'eye' thingy

alexander-schmidt:
Peek 2022-06-29 12-24

- The eye is now a regular button, with regular hover interactions.
- The button is however somewhat special:
  - Semi-transparent if password is hidden; opaque if password is shown.
  - Icon size is artificially increased via CSS (to match the effective size before this PR).
- Added button titles (depending on the state).
- Changed some apparently unnecessary API methods to `private`.


oliver-richers (approved):
nice

PR #337, opened 2022-06-29
PLAT-982 Remove DBMS configuration table

alexander-schmidt:


oliver-richers (approved):
the best database table is no database table

PR #338, opened 2022-06-29
PLAT-986 Use consistent paging icons for PDF and image previews

alexander-schmidt:
Peek 2022-06-29 13-16


oliver-richers (approved):
nice

PR #344, opened 2022-07-01
PLAT-999 Fix Handling of not-accessible Page

daniel-klose:
- When opening forspace with a valid page parameter but not having the rights to see the page, or the page doesnt exist, the navigation has broken display and missing styles.
  - Added a fallback to start page in that case with an alert that informs the user


oliver-richers (approved):
nice

PR #367, opened 2022-07-04
PLAT-898 Upgrade bcprov to 1.70

alexander-schmidt:


oliver-richers (approved):
if it compiles, it's fine

PR #375, opened 2022-07-07
PLAT-949-2 Implemented New DomAutoCompleteDefaultInputEngine

oliver-richers:
- this PR adds the new `DomAutoCompleteDefaultInputEngine` which has proper deduplication of display strings
- for now, only `EmfEnumTableRowInput` and `EmfForeignIndirectEntityAttribute` use this new engine
- `DomAutoCompleteDefaultInputEngine` aims to become the basis for **all** in-memory auto-complete engines in future PRs

### EmfEnumTableRowInput

image
image

### EmfForeignIndirectEntityAttribute
image


platform-dom/src/main/java/com/softicar/platform/dom/elements/input/auto/DomAutoCompleteDefaultInputEngine.java
```diff
@@ -0,0 +1,208 @@
...
+ * By default, the {@link IDisplayString} for each element is determined by
+ * testing if the element implements the interface {@link IDisplayable}. If this
+ * fails, the {@link Object#toString} method is used as a fallback. This
+ * mechanism can be overridden by calling {@link #setDisplayFunction(Function)}.
+ * <p>
+ * Elements evaluating to equal {@link IDisplayString} instances are
+ * deduplicated by appending appropriate suffixes. The comparison is done
+ * case-insensitive, although the letter cases of the respective
+ * {@link IDisplayString} instances are retained, e.g. "foo" and "Foo" will
+ * become "foo (1)" and "Foo (2)".
```
alexander-schmidt, line 37: I think, in addition to being case-insensitive, the comparison also needs to be diacritic-insensitive.
Let there be two entities, a and b, with the following display names:
- a: "Bar"
- b: "Bǎr" 

That's a https://en.wikipedia.org/wiki/Caron diacritic. Now, try to select b with an English or German keyboard layout. Yeah...
Btw, we already have `DiacriticNormalizer` for that.

platform-dom/src/main/java/com/softicar/platform/dom/elements/input/auto/DomAutoCompleteDefaultInputEngine.java
```diff
@@ -0,0 +1,208 @@
...
+ * <p>
+ * To ensure a deterministic order for deduplicated {@link IDisplayString}, the
+ * elements should implement {@link Comparable}. This order can be customized by
+ * calling {@link #setComparator}.
+ * <p>
+ * The comparator supplied to {@link #setComparator} will <b>not</b> be used to
+ * order the elements returned by {@link #findMatches}. Instead,
+ * {@link #findMatches} always returns elements order by their translated
+ * {@link IDisplayString} in a lexicographical manner. Thus, the order depends
```
alexander-schmidt, line 46:
```suggestion
 * {@link IDisplayString}, in a lexicographical manner. Thus, the order depends
```

platform-dom/src/main/java/com/softicar/platform/dom/elements/input/auto/DomAutoCompleteDefaultInputEngine.java
```diff
@@ -0,0 +1,208 @@
...
+
+	private class Cache {
+
+		private final Map<String, T> stringToElementMap;
+		private final Map<T, String> elementToStringMap;
+
+		public Cache() {
+
+			this.stringToElementMap = new DomAutoCompleteDisplayStringDeduplicator<>(displayFunction, comparator).apply(loader.get());
+			this.elementToStringMap = stringToElementMap.entrySet().stream().collect(Collectors.toMap(Entry::getValue, Entry::getKey));
```
alexander-schmidt, line 159: Using `Collectors.toMap` might be dangerous here. Quoting from its Javadoc:
> There are no guarantees on the type, [...] of the {@code Map} returned.

So, an Exception could occur if, for example, `T` is non-`Comparable` (it's unbounded, after all) and a `TreeMap` is created.

platform-dom/src/test/java/com/softicar/platform/dom/elements/input/auto/AbstractDomAutoCompleteDefaultInputEngineTest.java
```diff
@@ -0,0 +1,74 @@
...
+import com.softicar.platform.common.core.i18n.IDisplayString;
+import com.softicar.platform.common.core.i18n.IDisplayable;
+import com.softicar.platform.common.testing.AbstractTest;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class AbstractDomAutoCompleteDefaultInputEngineTest extends AbstractTest {
```
alexander-schmidt, line 12: should actually be `abstract`

alexander-schmidt (changes requested):

platform-dom/src/main/java/com/softicar/platform/dom/elements/input/auto/DomAutoCompleteDefaultInputEngine.java
oliver-richers: done

platform-dom/src/main/java/com/softicar/platform/dom/elements/input/auto/DomAutoCompleteDefaultInputEngine.java
oliver-richers: - unlikely, i think that will break lots of code if they ever change that to TreeMap
- anyway, now doing it manually, so sad

platform-dom/src/test/java/com/softicar/platform/dom/elements/input/auto/AbstractDomAutoCompleteDefaultInputEngineTest.java
oliver-richers: fixed

platform-dom/src/main/java/com/softicar/platform/dom/elements/input/auto/DomAutoCompleteDefaultInputEngine.java
alexander-schmidt: - yes
- yes
- yes

alexander-schmidt (approved):

platform-dom/src/main/java/com/softicar/platform/dom/elements/input/auto/DomAutoCompleteDefaultInputEngine.java
oliver-richers: ok PLAT-1032 created (and related PLAT-1033)

PR #378, opened 2022-07-07
PLAT-958 Prioritise primary file server at file lookup

daniel-klose:


platform-core-module/src/main/java/com/softicar/platform/core/module/file/stored/server/AGStoredFileServer.java
```diff
@@ -13,6 +15,20 @@ public IDisplayString toDisplayWithoutId() {
 		return IDisplayString.create(getUrl());
 	}
 
+	public static List<AGStoredFileServer> getAllActiveWithPrimaryFirst() {
+
+		List<AGStoredFileServer> servers = new ArrayList<>();
+		Optional<AGStoredFileServer> primaryServer = getInstance();
+		getInstance().ifPresent(servers::add);
+		AGStoredFileServer.TABLE//
```
alexander-schmidt, line 23: Could be expressed as a single statement, via SQML. Probably not worth the hassle though. Just sayin.

alexander-schmidt (changes requested):

platform-core-module/src/main/java/com/softicar/platform/core/module/file/stored/server/AGStoredFileServer.java
```diff
@@ -13,6 +15,20 @@ public IDisplayString toDisplayWithoutId() {
 		return IDisplayString.create(getUrl());
 	}
 
+	public static List<AGStoredFileServer> getAllActiveWithPrimaryFirst() {
```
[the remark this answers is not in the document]
daniel-klose: added validator to prevent the primary server from being deactivated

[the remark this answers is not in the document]
alexander-schmidt: I think this can be circumvented by first deactivating a file server, and then adding it as primary. Solving this via validators appears quite indirect to me. I think we should instead change `getPrimary` to return empty if the primary file server is disabled. Less code, consistent behavior.

platform-core-module/src/main/java/com/softicar/platform/core/module/file/stored/server/AGStoredFileServer.java
```diff
@@ -13,7 +15,28 @@ public IDisplayString toDisplayWithoutId() {
 		return IDisplayString.create(getUrl());
 	}
 
-	public static Optional<AGStoredFileServer> getInstance() {
+	public boolean isPrimaryServer() {
```
alexander-schmidt, line 18: If the method below is called `getPrimary`, we should omit the tailing `Server` here as well (or add it to both method names).

alexander-schmidt (changes requested):

platform-core-module/src/main/java/com/softicar/platform/core/module/file/stored/server/AGStoredFileServer.java
```diff
@@ -13,10 +15,32 @@ public IDisplayString toDisplayWithoutId() {
...
+		return servers;
+	}
+
+	public static Optional<AGStoredFileServer> getPrimary() {
 
 		return Optional//
 			.ofNullable(AGCoreModuleInstance.getInstance())
-			.map(AGCoreModuleInstance::getPrimaryFileServer);
+			.map(AGCoreModuleInstance::getPrimaryFileServer)
+			.map(it -> it.isActive()? it : null);
```
alexander-schmidt, line 44:
```suggestion
			.filter(AGStoredFileServer::isActive);
```

alexander-schmidt (changes requested):

PR #384, opened 2022-07-11
PLAT-977-2 Refactor DockerContainerController

alexander-schmidt:
- Implemented fully flexible support for run-options (i.e. options passed to the `docker run` command).
- Can now handle containers which don't provide health state.


oliver-richers (approved):
nice

PR #393, opened 2022-07-19
PLAT-1052 Define date format via locale

alexander-schmidt:
image

### DIFF
--- a/platform-core-module/src/main/java/com/softicar/platform/core/module/localization/AGLocalization.sqml
+++ b/platform-core-module/src/main/java/com/softicar/platform/core/module/localization/AGLocalization.sqml
@@ -9,6 +9,7 @@ TABLE AGLocalization 'Core.Localization' {
 	AGCoreLanguage language
 	String decimalSeparator [MAXLENGTH=255]
 	String digitGroupSeparator = '' [MAXLENGTH=255]
+	String dateFormat [MAXLENGTH=255]
 
 	UK name
 }
--- a/platform-core-module/src/main/java/com/softicar/platform/core/module/localization/LocalizationValidator.java
+++ b/platform-core-module/src/main/java/com/softicar/platform/core/module/localization/LocalizationValidator.java
@@ -2,9 +2,9 @@ package com.softicar.platform.core.module.localization;
 
 import com.softicar.platform.common.core.CommonCoreI18n;
 import com.softicar.platform.common.core.utils.DevNull;
-import com.softicar.platform.common.date.Day;
 import com.softicar.platform.emf.validation.AbstractEmfValidator;
 import java.text.SimpleDateFormat;
+import java.util.Date;
 
 public class LocalizationValidator extends AbstractEmfValidator<AGLocalization> {
 
@@ -73,9 +73,8 @@ public class LocalizationValidator extends AbstractEmfValidator<AGLocalization> 
 			return true;
 		}
 
-		var date = Day.today().toDate();
 		try {
-			new SimpleDateFormat(dateFormat).format(date);
+			new SimpleDateFormat(dateFormat).format(new Date());
 			return false;
 		} catch (Exception exception) {
 			DevNull.swallow(exception);
--- a/platform-core-module/src/main/java/com/softicar/platform/core/module/localization/LocalizationValidator.java
+++ b/platform-core-module/src/main/java/com/softicar/platform/core/module/localization/LocalizationValidator.java
@@ -1,8 +1,10 @@
 package com.softicar.platform.core.module.localization;
 
 import com.softicar.platform.common.core.CommonCoreI18n;
-import com.softicar.platform.common.core.exceptions.SofticarUserException;
+import com.softicar.platform.common.core.utils.DevNull;
+import com.softicar.platform.common.date.Day;
 import com.softicar.platform.emf.validation.AbstractEmfValidator;
+import java.text.SimpleDateFormat;
 
 public class LocalizationValidator extends AbstractEmfValidator<AGLocalization> {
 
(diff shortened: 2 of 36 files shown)


platform-core-module/src/main/java/com/softicar/platform/core/module/page/service/login/PageServiceLoginDiv.java
```diff
@@ -168,7 +168,7 @@ private void appendMaintenanceInProgressInfo(AGMaintenanceWindow maintenanceWind
 						DomMessageType.WARNING,
 						CoreI18n.MAINTENANCE_IS_CURRENTLY_IN_PROGRESS
 							.concat("\n")
-							.concat(CoreI18n.EXPECTED_END_ARG1.toDisplay(maintenanceWindow.getExpectedEnd().toHumanString()))));
+							.concat(CoreI18n.EXPECTED_END_ARG1.toDisplay(maintenanceWindow.getExpectedEnd().toDisplay()))));
```
oliver-richers, line 171: I think the `toDisplay` call should not be necessary, and I assume many developer will even forget about it an the locale will not be effective -> defect

platform-core-module/src/main/java/com/softicar/platform/core/module/program/execution/AGProgramExecutionTable.java
```diff
@@ -34,11 +34,11 @@ public void customizeAttributeProperties(IEmfAttributeList<AGProgramExecution> a
 
 		attributes//
 			.editAttribute(AGProgramExecution.STARTED_AT)
-			.setDisplayFactory(value -> new EmfDayTimeDisplay(value, DayTime::toString));
+			.setDisplayFactory(value -> new EmfDayTimeDisplay(value, DayTime::toDisplay));
```
oliver-richers, line 37: - I am afraid that may developers will forget to call `toDisplay` instead of `toString`.
- are we sure `toString` should not respect `CurrentLocale`?

platform-core-module/src/main/java/com/softicar/platform/core/module/test/fixture/CoreModuleTestFixtureMethods.java
```diff
@@ -75,24 +75,26 @@ default AGUserPassword insertPassword(AGUser user, String password) {
...
 			.save();
 	}
 
 	default AGLocalization insertLocalizationPresetGermany() {
 
-		return insertLocalizationPreset("Deutschland", LanguageEnum.GERMAN, ",", ".");
+		return insertLocalizationPreset("Deutschland", LanguageEnum.GERMAN, ",", ".", "dd.MM.yyyy");
 	}
 
 	default AGLocalization insertLocalizationPresetUsa() {
```
oliver-richers, line 95: - actually `yyyy-MM-dd` is not USA, that would be rather `MM/dd/yyyy`
- we might also rename this method to `insertLocalizationPresetInterantional` and the locale `International`

platform-core-module/src/main/java/com/softicar/platform/core/module/localization/AGLocalization.sqml
```diff
@@ -9,6 +9,7 @@ TABLE AGLocalization 'Core.Localization' {
 	AGCoreLanguage language
 	String decimalSeparator [MAXLENGTH=255]
 	String digitGroupSeparator = '' [MAXLENGTH=255]
+	String dateFormat [MAXLENGTH=255]
```
oliver-richers, line 12: I see no adaption of the LocalizationValidator and LocalizationValidatorTest

oliver-richers (changes requested):

platform-core-module/src/main/java/com/softicar/platform/core/module/program/execution/AGProgramExecutionTable.java
alexander-schmidt: In the long run, I think `toString` should indeed respect `CurrentLocale`, and provide output that is formatted accordingly - I see no way around this. Yet, I considered that too dangerous for now:
Before changing the output of `toString`, we must somehow ensure that we currently _never_ use `toString` when we in fact need an ISO-formatted date (that's precisely why it's a bad practice to rely on the output of `toString` in a non-user-facing context, btw).
As long as we don't find an efficient way to detect this, we're kind of stuck in the current middle ground, in which locale-aware date formatting is _possible_ (via `toDisplay`) while unfortunately not being the default behavior.

platform-core-module/src/main/java/com/softicar/platform/core/module/page/service/login/PageServiceLoginDiv.java
alexander-schmidt: see the thread on `AGProgramExecutionTable`. closing this one.

platform-core-module/src/main/java/com/softicar/platform/core/module/test/fixture/CoreModuleTestFixtureMethods.java
alexander-schmidt: changed to `International`... I think nobody should ever use that nonsensical `MM/dd/yyyy` format

platform-core-module/src/main/java/com/softicar/platform/core/module/localization/AGLocalization.sqml
alexander-schmidt: nice catch! forgot about those.

alexander-schmidt: ...`LocalizationValidator` was wrongfully implemented: It threw exceptions instead of adding errors to the validation results. This led to premature termination of the overall validation when trying to create or edit a record from the UI, where fixing one input error could unlock further input errors.
`LocalizationValidator` was fixed accordingly, which entailed a rewrite of `LocalizationValidatorTest`.

alexander-schmidt: image

oliver-richers: nice

platform-core-module/src/main/java/com/softicar/platform/core/module/localization/LocalizationValidator.java
```diff
@@ -57,4 +66,20 @@ private boolean containsIllegalCharacters(String text) {
...
+
+	private boolean isInvalidDateFormat(String dateFormat) {
+
+		if (dateFormat.isBlank()) {
+			return true;
+		}
+
+		var date = Day.today().toDate();
+		try {
+			new SimpleDateFormat(dateFormat).format(date);
```
oliver-richers, line 78:
```suggestion
		try {
			new SimpleDateFormat(dateFormat).format(new Date());
```

platform-core-module/src/main/java/com/softicar/platform/core/module/page/service/login/PageServiceLoginDiv.java
oliver-richers: - still, I think `toDisplay()` must not be necessary
- see `I18nFormatArgumentsLocalizer` how we do that for numbers, also see comment below

platform-core-module/src/main/java/com/softicar/platform/core/module/program/execution/AGProgramExecutionTable.java
oliver-richers: - if we don't change the default behavior of `toString` now, we will never change it ;) but see next point
- I agree that changing `toString` is dangerous, but then we should at least ensure that anything related to `IDisplayString` will always use `toDisplay` and not `toString`, again, see `I18nFormatArgumentsLocalizer` and consider adding an `if (argument instanceof IDisplayable)` case

platform-core-module/src/test/java/com/softicar/platform/core/module/localization/LocalizationValidatorTest.java
```diff
@@ -1,58 +1,90 @@
...
 import com.softicar.platform.db.runtime.test.AbstractDbTest;
-import com.softicar.platform.emf.validation.result.EmfValidationResult;
 import java.util.List;
 import org.junit.Test;
 
 public class LocalizationValidatorTest extends AbstractDbTest {
 
+	public LocalizationValidatorTest() {
+
+		// TODO this should not be necessary
```
oliver-richers, line 14: please elaborate? why then is it necessary? ;)

oliver-richers (changes requested):

platform-core-module/src/main/java/com/softicar/platform/core/module/test/fixture/CoreModuleTestFixtureMethods.java
oliver-richers: LOL, so are imperial units like square feet, still they are used :D

platform-core-module/src/main/java/com/softicar/platform/core/module/program/execution/AGProgramExecutionTable.java
alexander-schmidt: Did the `I18nFormatArgumentsLocalizer` thing. We chould change the behavior of `toString` - but I think we should do so in a separate PR, so that it's easily reversible (without taking away the feature).

oliver-richers (approved):

oliver-richers (approved):
ok

PR #398, opened 2022-07-21
PLAT-1053 Fix visualization when simultaneously using row selection and an action column handler

alexander-schmidt:
image
----
image
----
image


platform-emf/src/main/java/com/softicar/platform/emf/data/table/EmfDataTableRow.java
```diff
@@ -125,33 +128,50 @@ private <T> void setDefaultTextAlign(IEmfDataTableColumn<R, T> column, EmfDataTa
 		cell.setStyle(cell.getColumn().getSettings().getAlignmentOrDefault(textAlign));
 	}
 
-	private class ActionCell extends DomCell implements IEmfDataTableActionCell<R> {
+	private class ActionCell extends DomDelegatingParentElement implements IEmfDataTableActionCell<R> {
```
oliver-richers, line 131: - is it really necessary to use `DomDelegatingParentElement` here?
- wouldn't it be much easier to drop the `container` and fix it all with CSS?

oliver-richers (changes requested):

platform-emf/src/main/java/com/softicar/platform/emf/data/table/EmfDataTableRow.java
alexander-schmidt: Without the container, we'd need to directly style the `td` element. That is, we'd need to align children next to each other, with a gap. `flex` or `grid` are the only reasonable means I know to achieve this (`padding` or `margin` are not viable, IMO).
Any suggestions?

oliver-richers: ok, seems there is currently no better solution

oliver-richers (approved):

PR #405, opened 2022-07-26
PLAT-1028 Added fix for orphaned ProgramExection records

daniel-klose:


platform-core-module/src/main/java/com/softicar/platform/core/module/program/execution/cleanup/ProgramExecutionsCleaner.java
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	private void terminateExecutionAndCreateEvent(AGProgramExecution execution) {
+
+		execution//
+			.setFailed(true)
+			.setTerminatedAt(DayTime.now())
+			.save();
+		new SystemEventBuilder(AGSystemEventSeverityEnum.ERROR, CoreI18n.PROGRAM_EXECUTION_FAILED.toString())//
+			.addProperty("program", Programs.getProgramName(execution.getProgramUuid().getUuid()).toString())
```
oliver-richers, line 34: this looks like code that would be repeated all over the place, suggesting to add method `AGProgramExecution.getProgram()` and then call `execution.getProgram().toDisplay()`here

platform-core-module/src/main/java/com/softicar/platform/core/module/program/execution/cleanup/ProgramExecutionsCleaner.java
```diff
@@ -0,0 +1,37 @@
...
+import com.softicar.platform.common.date.DayTime;
+import com.softicar.platform.core.module.CoreI18n;
+import com.softicar.platform.core.module.event.SystemEventBuilder;
+import com.softicar.platform.core.module.event.severity.AGSystemEventSeverityEnum;
+import com.softicar.platform.core.module.program.Programs;
+import com.softicar.platform.core.module.program.execution.AGProgramExecution;
+import com.softicar.platform.core.module.program.state.AGProgramState;
+import com.softicar.platform.db.core.transaction.DbTransaction;
+
+public class ProgramExecutionsCleaner {
```
oliver-richers, line 12: there should be a `ProgramExecutionsCleanerTest`

platform-core-module/src/main/java/com/softicar/platform/core/module/program/execution/cleanup/ProgramExecutionsCleaner.java
```diff
@@ -0,0 +1,37 @@
...
+		}
+	}
+
+	private void terminateExecutionAndCreateEvent(AGProgramExecution execution) {
+
+		execution//
+			.setFailed(true)
+			.setTerminatedAt(DayTime.now())
+			.save();
+		new SystemEventBuilder(AGSystemEventSeverityEnum.ERROR, CoreI18n.PROGRAM_EXECUTION_FAILED.toString())//
```
oliver-richers, line 33: I wonder if this shouldn't be rather `PROGRAM_EXECUTION_WAS_ABORTED` or similar?

oliver-richers (changes requested):

platform-core-module/src/main/java/com/softicar/platform/core/module/program/execution/cleanup/ProgramExecutionsCleaner.java
daniel-klose: Hm that implies that all orphaned executions were aborted, dont know if that is correct.
Failed is imho a good term, because a execution had to be cleaned, therefore it failed in some way

platform-core-module/src/main/java/com/softicar/platform/core/module/program/execution/cleanup/ProgramExecutionsCleaner.java
daniel-klose: ohhh

platform-core-module/src/main/java/com/softicar/platform/core/module/program/execution/cleanup/ProgramExecutionsCleaner.java
oliver-richers: actually, if the execution record remained, that can only mean that someone killed the JVM, so the program was actually aborted. In my book, "fail" means there was some kind of error in the program execution, with a high likeliness of reoccurring, which is here not given.

but if you insist, this will not stop me to accept

oliver-richers: maybe we should even be more explicit and write something like ORPHANED_PROGRAM_EXECUTION, so the system admin may remember: "Ah these executions where kill the yesterday, when I restarted the server."

oliver-richers (changes requested):

platform-core-module/src/main/java/com/softicar/platform/core/module/program/execution/cleanup/ProgramExecutionsCleaner.java
daniel-klose: oh thats a good idea yeah

daniel-klose: also changed system event to be a warning, because if it really only happends when the JVM gets restartet, its not exactly an error in my book

oliver-richers (approved):
nice

platform-core-module/src/main/java/com/softicar/platform/core/module/program/execution/cleanup/ProgramExecutionsCleaner.java
oliver-richers: ok

PR #406, opened 2022-07-26
PLAT-1043 Add scope field to emf form

daniel-klose:
- Admin User with Permissions:
image
  - Normal User without Permission:
image


platform-emf/src/main/java/com/softicar/platform/emf/form/EmfFormAttributesDiv.java
```diff
@@ -20,4 +22,22 @@ public EmfFormAttributesDiv(R tableRow, boolean editMode) {
 			}
 		}
 	}
+
+	//FIXME This works, but needs simplification
+	private boolean isScopeViewPermissionMissing(IEmfAttribute<R, ?> attribute) {
```
oliver-richers, line 27: I think we should rather show the attribute in any case, but then disable the button to open the view, i think this will be less confusing for users and easier to implement. it will also be rarely the case that someone cannot see the scope of an entity that they can see

platform-emf/src/main/java/com/softicar/platform/emf/management/EmfManagementDataTable.java
```diff
@@ -141,7 +141,7 @@ void addOrderBy(EmfDataTableDivBuilder<R> builder, IEmfAttribute<R, ?> attribute
 
 	private <V> boolean showAttribute(IEmfAttribute<R, V> attribute) {
 
-		if (attribute.isConcealed()) {
+		if (attribute.isConcealed() || entityTable.getScopeAttribute().map(it -> it.equals(attribute)).orElse(false)) {
```
oliver-richers, line 144: we should extract this to a IEmfAttribute.isScope method

platform-emf/src/main/java/com/softicar/platform/emf/management/EmfManagementDiv.java
```diff
@@ -223,6 +224,14 @@ private boolean isNullable(IEmfAttribute<R, ?> attribute) {
 			return false;
 		}
 
+		private boolean isScope(IEmfAttribute<R, ?> attribute) {
```
oliver-richers, line 227: see above

oliver-richers (changes requested):

platform-emf/src/main/java/com/softicar/platform/emf/form/EmfFormAttributesDiv.java
daniel-klose: well the button already isnt appended if the user cant access it, see screenshot

platform-emf/src/main/java/com/softicar/platform/emf/attribute/EmfInheritedAttribute.java
```diff
@@ -98,6 +98,12 @@ public IStaticObject getTestMarker() {
 		return originalAttribute.getTestMarker();
 	}
 
+	@Override
+	public boolean isScope() {
+
+		return getTable().getScopeField().map(it -> it.equals(baseField)).orElse(false);
```
oliver-richers, line 104: uh, that's wrong, strange that this compiles, it must be
```suggestion
		return originalAttribute.isScope();
```

platform-emf/src/main/java/com/softicar/platform/emf/attribute/IEmfAttribute.java
```diff
@@ -97,6 +97,14 @@
 	 */
 	IStaticObject getTestMarker();
 
+	/**
+	 * Returns true if this attribute is the scope, false otherwise.
+	 *
+	 * @return true if this is the scope attribute, false otherwise
+	 */
+
```
oliver-richers, line 105:
```suggestion
```

platform-emf/src/main/java/com/softicar/platform/emf/attribute/field/EmfFieldAttribute.java
```diff
@@ -107,10 +109,20 @@ public void setValue(R tableRow, V value) {
...
 	}
 
+	@Override
+	public boolean isScope() {
+
+		if (field instanceof IDbForeignRowField) {
+			IDbForeignRowField<R, ?, ?> scopeField = CastUtils.cast(field);
+			return getTable().getScopeField().map(it -> it.equals(scopeField)).orElse(false);
+		}
+		return false;
```
oliver-richers, line 119: can be simplified
```suggestion
		return getTable().isScopeAttribute(this);
```

oliver-richers (changes requested):

oliver-richers (approved):

PR #423, opened 2022-08-15
PLAT-1074 Fix previews for PDFs with certain embedded JPEGs

alexander-schmidt:
- Added dependency on `imageio-jpeg`.
- Fixes "Numbers of source Raster bands and source color space components do not match" when trying to render certain JPEGs.
- See https://stackoverflow.com/questions/18351583/illegalargumentexception-numbers-of-source-raster-bands-and-source-color-space/18400559#18400559

image


oliver-richers (dismissed):
great!

oliver-richers (approved):

PR #440, opened 2022-08-16
PLAT-1082 Aggregate marker enums per project

daniel-klose:


platform-dom/src/main/java/com/softicar/platform/dom/DomTestMarker.java
```diff
@@ -5,7 +5,33 @@
 public enum DomTestMarker implements IStaticObject {
```
oliver-richers, line 5: shouldn't we sort alphabetically? here and in the other enums

oliver-richers (changes requested):

platform-dom/src/main/java/com/softicar/platform/dom/DomTestMarker.java
daniel-klose: already did that, seems I forgot to sort some entries though

oliver-richers (approved):

PR #446, opened 2022-08-18
PLAT-827 Implemented Proper Waiting for Server in Selenium Tests

oliver-richers:


platform-ajax/src/main/java/com/softicar/platform/ajax/testing/selenium/engine/common/AjaxSeleniumTestServerWaiter.java
```diff
@@ -0,0 +1,65 @@
...
+			this.waitScriptKey = getJavascriptExecutor().pin(WAIT_SCRIPT);
+		}
+		return waitScriptKey;
+	}
+
+	private static String createWaitScript() {
+
+		return new StringBuilder()//
+			.append("%s = false;".formatted(FINISHED_VARIABLE))
+			.append("function waitForServer() {;")
```
alexander-schmidt, line 55: That tailing semicolon looks odd. Why does this even work?

oliver-richers: indeed ;) well it works, since it's just a redundant empty statement

alexander-schmidt: oh. yea makes sense.

platform-ajax/src/test/java/com/softicar/platform/ajax/testing/selenium/engine/common/AjaxSeleniumTestServerWaiterTest.java
```diff
@@ -0,0 +1,84 @@
...
+		this.testDiv = openTestNode(TestDiv::new);
+		this.events = new ArrayList<>();
+	}
+
+	@Test
+	public void test() {
+
+		send(testDiv.input1, "a");
+		click(testDiv.input2);
+		waitForServer();
```
alexander-schmidt, line 31: Shouldn't we test `AjaxSeleniumTestServerWaiter` here directly? Like `new AjaxSeleniumTestServerWaiter(webDriver).waitForServer()`. Otherwise, we kind of assume an implementation detail of `AbstractAjaxSeleniumLowLevelTest.waitForServer`.

platform-ajax/src/test/java/com/softicar/platform/ajax/testing/selenium/engine/common/AjaxSeleniumTestServerWaiterTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+		assertEquals("[FOCUS#1, INPUT#1, BLUR#1, FOCUS#2, CLICK#2]", events.toString());
```
alexander-schmidt, line 33: I don't think the test setup and this assertion in particular are sufficient. With a certain amount of luck, this assertion could pass even without the `waitForServer` call.
Could we somehow provoke failure, such that this assertion is highly likely to fail if the `waitForServer` call would be omitted? e.g. via introduction of an artificial delay

alexander-schmidt (changes requested):

platform-ajax/src/test/java/com/softicar/platform/ajax/testing/selenium/engine/common/AjaxSeleniumTestServerWaiterTest.java
oliver-richers: changed

platform-ajax/src/test/java/com/softicar/platform/ajax/testing/selenium/engine/common/AjaxSeleniumTestServerWaiterTest.java
oliver-richers: did so

alexander-schmidt (approved):

PR #450, opened 2022-08-23
PLAT-1080 Reworked CorePermissions

daniel-klose:


platform-core-module/src/main/java/com/softicar/platform/core/module/CoreModule.java
```diff
@@ -47,4 +49,20 @@ public static EmfPagePath getParentPagePath() {
 
 		return new EmfPagePath().append(PARENT_FOLDER_TITLE);
 	}
+
+	//FIXME Should be removed when PLAT-1099 is done
```
oliver-richers, line 53: this is rather a TODO, there is no real problem with having there methods
```suggestion
	// TODO should be removed with PLAT-1099
```

platform-core-module/src/main/java/com/softicar/platform/core/module/CorePermissions.java
```diff
@@ -1,15 +1,11 @@
...
-			new CoreModulePermission("abeb1897-5079-4bf8-9f1b-6c7cd7e71890", CoreI18n.SUPER_USER);
-
-	final CoreModulePermission ACCESS_MANAGEMENT =//
-			new CoreModulePermission("c0b0e2ea-7475-4e39-addf-f977f5eb8986", CoreI18n.ACCESS_MANAGEMENT);
-
-	final CoreModulePermission SYSTEM_ADMINISTRATION =//
-			new CoreModulePermission("e1bb654c-4a85-4c2f-b0a2-1116b2fca657", CoreI18n.SYSTEM_ADMINISTRATION);
+	IEmfModulePermission<AGCoreModuleInstance> VIEW = EmfDefaultModulePermissions.getModuleView();
+	IEmfModulePermission<AGCoreModuleInstance> OPERATION = EmfDefaultModulePermissions.getModuleOperation();
+	IEmfModulePermission<AGCoreModuleInstance> ADMINISTRATION = EmfDefaultModulePermissions.getModuleAdministation();
```
oliver-richers, line 10: typo
```suggestion
	IEmfModulePermission<AGCoreModuleInstance> ADMINISTRATION = EmfDefaultModulePermissions.getModuleAdministration();
```

platform-core-module/src/main/java/com/softicar/platform/core/module/ajax/session/SofticarAjaxSession.java
```diff
@@ -32,9 +32,9 @@ public static Optional<SofticarAjaxSession> getInstance(HttpSession session) {
 		return new AjaxSessionAttributeManager(session).getInstance(SofticarAjaxSession.class);
 	}
 
-	public boolean isSuperUser() {
+	public boolean isCoreModuleAdmin() {
 
-		return user.get().isSuperUser();
+		return user.get().isCoreModuleAdmin();
```
oliver-richers, line 37: the abbreviation `Admin` is arbitrary and shouldn't we use same term as before, that is, `isAdminitrator()`

platform-emf/src/main/java/com/softicar/platform/emf/management/EmfManagementDiv.java
```diff
@@ -236,7 +236,7 @@ private boolean isCreationAllowed() {
 			if (scopeEntity != null) {
 				return entityTable.isCreationPossible(scopeEntity) && entityTable.isCreationAllowed(scopeEntity);
 			} else {
-				return CurrentBasicUser.get().isSuperUser();
+				return false;
```
oliver-richers, line 239: is this correct? can admins still create users with this change?

platform-core-module/src/main/java/com/softicar/platform/core/module/user/session/UserSessionManager.java
```diff
@@ -40,7 +41,7 @@ private void invalidateSessionIfNotAdministrator(HttpSession session) {
 
 	private void checkUserAndInvalidateIfNotAdministrator(SofticarAjaxSession ajaxSession, HttpSession session) {
 
-		if (!CorePermissions.SYSTEM_ADMINISTRATION.test(ajaxSession.getUser())) {
+		if (!CorePermissions.ADMINISTRATION.test(AGCoreModuleInstance.getInstance(), ajaxSession.getUser())) {
```
oliver-richers, line 44: cant we just use `user.isAdministrator()` here?

platform-core-module/src/main/java/com/softicar/platform/core/module/user/AGUser.java
```diff
@@ -69,10 +69,9 @@ public static AGUser getSystemUser() {
 		return AGCoreModuleInstance.getInstance().getSystemUser();
 	}
 
-	@Override
-	public Boolean isSuperUser() {
+	public Boolean isCoreModuleAdmin() {
```
oliver-richers, line 72: isAdministrator()?

platform-core-module/src/main/java/com/softicar/platform/core/module/start/page/StartPageDiv.java
```diff
@@ -39,7 +39,7 @@ public StartPageDiv() {
 
 	private void addPendingSystemEventsSection(Collection<IDomElement> sections) {
 
-		if (CorePermissions.SYSTEM_ADMINISTRATION.test(CurrentUser.get())) {
+		if (CorePermissions.ADMINISTRATION.test(AGCoreModuleInstance.getInstance(), CurrentUser.get())) {
```
oliver-richers, line 42:
```suggestion
		if (CurrentUser.get().isAdministrator()) {
```
please look for similar places to simplify

oliver-richers (changes requested):

platform-core-module/src/main/java/com/softicar/platform/core/module/user/AGUser.java
alexander-schmidt: **TL;DR:**
I think we _must_ inline `isCoreModuleAdmin` from both `AGUser` and `SofticarAjaxSession`.

**Long version:**
I think this method should not exist here in the first place. The same goes for the equally-named method on `SofticarAjaxSession`. Those feel out of place, and they're like nudging towards a cliff.
Payload software will handle `AGUser` objects all over the place; so `isCoreModuleAdmin`/`isAdministrator` might be called because it appears (!) to be a convenient way to check for administrative permissions.
But it's not. Because it's the wrong way to go; beyond the core module, that is. No other module should ever check for `CorePermissions.ADMINISTRATION`, so we should not make that check readily-available (on `AGUser`).
Also, we got a terminological mess here. If anything, it should be `hasAdministration`, instead of `isAdministrator` - because there is no such thing as a hard-coded "administrator" role in the core module (anymore). That mess can be avoided altogether - see the TL;DR above.

platform-emf/src/main/java/com/softicar/platform/emf/management/EmfManagementDiv.java
daniel-klose: didnt you say yesterday to just return false here?
And what does this have to do with users?

daniel-klose: but yes, I can still create users when I log into the admin user

platform-ajax/src/main/java/com/softicar/platform/ajax/customization/IAjaxStrategy.java
```diff
@@ -76,17 +76,17 @@ public interface IAjaxStrategy {
...
 	 * sessions.
 	 *
 	 * @param session
 	 *            the {@link HttpSession} to test (never <i>null</i>)
-	 * @return <i>true</i> if the given {@link HttpSession} is a super-user
+	 * @return <i>true</i> if the given {@link HttpSession} is a administrator
 	 *         session; <i>false</i> otherwise
 	 */
-	boolean isSuperUser(HttpSession session);
+	boolean isAdministrator(HttpSession session);
```
alexander-schmidt, line 89: In order to not wrongfully hint to a (nonexistent) "administrator" role, we could rename this method to `isAdministrative`. So if `true`, we got an "administrative session". Seems kind of natural.

platform-ajax/src/main/java/com/softicar/platform/ajax/request/AjaxRequest.java
```diff
@@ -199,8 +199,8 @@ public boolean isDeflateSupported() {
 		return acceptEncoding != null && acceptEncoding.toLowerCase().contains("deflate");
 	}
 
-	private boolean isSuperUser() {
+	private boolean isAdministrator() {
```
alexander-schmidt, line 202: see above

platform-core-module/src/main/java/com/softicar/platform/core/module/AGCoreModuleInstance.java
```diff
@@ -30,6 +29,11 @@ public static String getSystemIdentifier() {
 		return AGCoreModuleInstance.getInstance().getSystemName();
 	}
 
+	public Boolean isAdministrator(AGUser user) {
```
alexander-schmidt, line 32: If (!) we retain this method, we should rename it to `hasAdministrationPermission`. However, recommend inlining it.

platform-core-module/src/main/java/com/softicar/platform/core/module/CoreI18n.java
```diff
@@ -174,6 +174,8 @@ public interface CoreI18n extends EmfI18n {
 		.de("Kern-Sprache");
 	I18n0 CORE_LANGUAGES = new I18n0("Core Languages")//
 		.de("Kern-Sprachen");
+	I18n0 CORE_MODULE_ADMINISTRATORS_CANNOT_BE_IMPERSONATED = new I18n0("Core module administrators cannot be impersonated.")//
```
alexander-schmidt, line 177: This terminology quires discussion

platform-core-module/src/main/java/com/softicar/platform/core/module/CoreModule.java
```diff
@@ -47,4 +49,20 @@ public static EmfPagePath getParentPagePath() {
 
 		return new EmfPagePath().append(PARENT_FOLDER_TITLE);
 	}
+
+	// TODO Remove with PLAT-1099
+	public static <T extends IEmfTableRow<T, ?>> IEmfPermission<T> getModuleAdministation() {
```
alexander-schmidt, line 54: Shouldn't those methods have a `Permission` suffix? Current naming appears odd to me.

platform-core-module/src/main/java/com/softicar/platform/core/module/CoreModuleMapper.java
```diff
@@ -0,0 +1,26 @@
+package com.softicar.platform.core.module;
+
+import com.softicar.platform.common.core.i18n.IDisplayString;
+import com.softicar.platform.emf.mapper.IEmfTableRowMapper;
+import com.softicar.platform.emf.table.row.IEmfTableRow;
+import java.util.Optional;
+
+class CoreModuleMapper<T extends IEmfTableRow<T, ?>> implements IEmfTableRowMapper<T, AGCoreModuleInstance> {
```
alexander-schmidt, line 8: I think this class deserves an explanatory sentence. Hard to grasp what/why this is.

platform-core-module/src/main/java/com/softicar/platform/core/module/CoreModule.java
daniel-klose: they get removed anyway
also, their terminology is the same as in `EmfDefaultModuleRoles`

platform-core-module/src/main/java/com/softicar/platform/core/module/CoreModuleMapper.java
daniel-klose: honestly I cant really think up a good explanation, Olli can best describe this I think :D

platform-core-module/src/main/java/com/softicar/platform/core/module/CoreI18n.java
alexander-schmidt: Btw, this functionality has been inaccessible for like 2yrs now. We should just delete it.

platform-core-module/src/main/java/com/softicar/platform/core/module/CoreModule.java
alexander-schmidt: As discussed, should be renamed to `getAdministationPermission`. Same goes for similar methods.

alexander-schmidt (approved):

platform-emf/src/main/java/com/softicar/platform/emf/management/EmfManagementDiv.java
```diff
@@ -59,7 +60,7 @@ public EmfManagementDiv(IEmfTable<R, P, S> entityTable, S scopeEntity) {
 	protected EmfManagementDiv(EmfManagementDivBuilder<R, P, S> builder) {
 
 		this.entityTable = builder.getEntityTable();
```
oliver-richers, line 62:
```suggestion
		this.entityTable = Objects.requireNonNull(builder.getEntityTable());
```

oliver-richers (dismissed):

alexander-schmidt (approved):

PR #452, opened 2022-08-24
PLAT-1104 Eliminate user impersonation feature

alexander-schmidt:


oliver-richers (approved):
the best feature is no feature

PR #466, opened 2022-08-31
PLAT-1124 Fix style of EmfFormFrame

daniel-klose:
Left: EmfFormFrame Right: EmfFormPopup
image


platform-emf/src/main/java/com/softicar/platform/emf/form/frame/EmfFormFrameHeader.java
```diff
@@ -14,7 +14,15 @@ public EmfFormFrameHeader() {
 	public EmfFormFrameHeader setCaption(IDisplayString entityTitle, IDisplayString operationTitle) {
 
 		removeChildren();
-		appendText(entityTitle.concat(" - ").concat(operationTitle).toString());
+
+		var captionDiv = appendChild(new DomDiv());
```
alexander-schmidt, line 18: I think we should have a `private class EmfFormFrameCaption` here - at least that's what we usually do when we have a special CSS class.

platform-emf/src/main/java/com/softicar/platform/emf/form/frame/EmfFormFrameHeader.java
```diff
@@ -14,7 +14,15 @@ public EmfFormFrameHeader() {
...
+		captionDiv.setCssClass(EmfCssClasses.EMF_FORM_FRAME_CAPTION);
+		captionDiv.appendText(entityTitle);
+
+		var subCaptionDiv = appendChild(new DomDiv());
```
alexander-schmidt, line 22: same as above

platform-emf/src/main/resources/com/softicar/platform/emf/emf-form-style.css
```diff
@@ -14,16 +14,33 @@
 	border-radius: var(--BORDER_RADIUS);
 	box-shadow: var(--BOX_SHADOW_OVERLAY);
 	display: inline-grid;
+	grid-auto-columns: max-content;
+	align-self: start;
 	outline-style: none;
+	gap: 0px;
```
alexander-schmidt, line 20: Why is this necessary, all of a sudden, and why won't things look gapless now?

platform-emf/src/main/resources/com/softicar/platform/emf/emf-form-style.css
```diff
@@ -78,9 +95,7 @@
 	grid-column: 1/3;
 }
 
-.EmfFormCommonActionsDiv {
-	gap: 4px;
-	display: flex;
+.DomActionBar.EmfFormCommonActionsDiv {
```
alexander-schmidt, line 98: Do handcrafted context menus still look like before now?

alexander-schmidt (changes requested):

platform-emf/src/main/java/com/softicar/platform/emf/form/frame/EmfFormFrameHeader.java
daniel-klose: hm, i looked in `DomPopup` and there is essentially the same thing as this, just a little bit different
image

platform-emf/src/main/resources/com/softicar/platform/emf/emf-form-style.css
daniel-klose: Well, EmfFormCommonActionsDiv extends DomActionBar, so nothing should have changed
If you mean handcrafted context menus like in the module classes page, then yes, everything is the same

platform-emf/src/main/resources/com/softicar/platform/emf/emf-form-style.css
daniel-klose: This is necessary to make the EmfFormFrame have the same dimensions and gaps as an EmfFormPopup.
EmfFormFrame has 2 Components, to the 5px gap would make a frame seem bigger. See screenshot
image

platform-emf/src/main/resources/com/softicar/platform/emf/emf-form-style.css
alexander-schmidt: Ooops, misread the class name. Never mind.

platform-emf/src/main/java/com/softicar/platform/emf/form/frame/EmfFormFrameHeader.java
alexander-schmidt: well, whatever

alexander-schmidt (approved):

PR #469, opened 2022-08-31
PLAT-1129 Remove redundant waitForServer calls

alexander-schmidt:


platform-ajax/src/test/java/com/softicar/platform/ajax/input/auto/complete/entity/AbstractAjaxAutoCompleteEntityTest.java
```diff
@@ -512,6 +512,8 @@ public class AssertionExecutor {
 
 			public void assertAll() {
 
+				AbstractAjaxAutoCompleteEntityTest.super.waitForServer();
```
alexander-schmidt, line 515: This is the only `waitForServer` call we still need.

platform-ajax/src/test/java/com/softicar/platform/ajax/input/auto/complete/entity/AbstractAjaxAutoCompleteEntityTest.java
```diff
@@ -658,17 +651,11 @@ public BodyProxy click() {
...
-
-		public BodyProxy waitForServer() {
-
-			AbstractAjaxAutoCompleteEntityTest.super.waitForServer();
-			return this;
-		}
 	}
 
-	protected class PopupProyx {
+	protected class PopupProxy {
```
alexander-schmidt, line 656: Fixed a random typo along the way.

PR #473, opened 2022-08-31
PLAT-1123 Improved PdfRenderer and Simplified DomImageViewer

oliver-richers:
- greatly simplified DomImageViewer
- refactored PdfToBufferedImageConverter into PdfRenderer
- refactored StoredFilePdfToZoomableImagesConverter into StoredFilePdfRenderer
- moved pdfbox dependencies from platform-core-module to platform-common-pdf
- implemented DomImageViewerTest
- implemented InMemoryResource
- implemented InMemoryImageResource
- extended AjaxSeleniumLowLevelTestEngineOutput


platform-core-module/src/main/java/com/softicar/platform/core/module/file/stored/preview/pdf/StoredFilePdfRenderer.java
```diff
@@ -0,0 +1,72 @@
...
+
+		Optional
+			.ofNullable(file.getContentType())
+			.filter(contentType -> contentType.contains("pdf"))
+			.orElseThrow(() -> new SofticarUserException(EmfI18n.THE_FILE_FORMAT_MUST_BE_PDF));
+
+		try (InputStream stream = file.getFileContentInputStream()) {
+			return renderPages(stream, file.getFileName());
+		} catch (Throwable throwable) {
+			Log.ferror(StackTraceFormatting.getStackTraceAsString(throwable));
```
alexander-schmidt, line 45: We should add a preamble here, like `"Caught an Exception while blablabla: "`, or `"Failed to blabla: "`.
Otherwise, in the log output, it would appear as if the Exception was uncaught.

platform-core-module/src/main/java/com/softicar/platform/core/module/file/stored/preview/pdf/StoredFilePdfPreviewPopup.java
```diff
@@ -25,19 +25,19 @@ public StoredFilePdfPreviewPopup(AGStoredFile file) {
...
 		}
 	}
 
-	private List<DomImageViewerImage> getPreviewImages(AGStoredFile file, ICssLength maxWidth) {
+	private List<IResource> getPreviewImages(AGStoredFile file) {
 
 		return Optional//
 			.of(file)
-			.map(it -> StoredFilePdfToZoomableImagesConverter.convertPagesToImages(file, maxWidth))
+			.map(it -> StoredFilePdfRenderer.renderPages(file))
```
alexander-schmidt, line 40: Odd lambda definition. `it` and `file` are the same. We should probably just use a method reference.

platform-ajax/src/main/java/com/softicar/platform/ajax/testing/selenium/engine/level/low/AjaxSeleniumLowLevelTestEngineOutput.java
```diff
@@ -229,4 +239,26 @@ public IDomModalPromptNodes<IDomNode> findModalPromptOrFail() {
...
+	// ------------------------------ CSS assert methods ------------------------------ //
+
+	public void assertCssClasses(Collection<ICssClass> expectedClasses, IDomNode node) {
+
+		var expectedValue = expectedClasses//
+			.stream()
+			.map(ICssClass::getName)
+			.sorted()
+			.collect(Collectors.joining(" "));
+		assertEquals(expectedValue, getAttributeValue(node, "class"));
```
alexander-schmidt, line 252: Can we rely on the stability of the order of the CSS classes, as returned by `getAttributeValue(node, "class")`? E.g. are they always ordered alphabetically?
If not, we should probably deal with that to avoid random failures... for instance, we could create two sets of CSS class names, and compare those sets.

alexander-schmidt (changes requested):

platform-core-module/src/main/java/com/softicar/platform/core/module/file/stored/preview/pdf/StoredFilePdfPreviewPopup.java
oliver-richers: not my code, I removed the complete method, cause it's useless

platform-core-module/src/main/java/com/softicar/platform/core/module/file/stored/preview/pdf/StoredFilePdfRenderer.java
oliver-richers: ok

platform-ajax/src/main/java/com/softicar/platform/ajax/testing/selenium/engine/level/low/AjaxSeleniumLowLevelTestEngineOutput.java
oliver-richers: no they are always sorted alphabetically but `DomElementCssClassAttributeManager`

alexander-schmidt (approved):

PR #480, opened 2022-09-05
PLAT-1136-2 Implement setPreventDefaultOnWheel

alexander-schmidt:
We currently cannot write automated tests for this. We could try https://www.selenium.dev/documentation/webdriver/actions_api/wheel/ but that would require a switch to Chrome based Selenium nodes. So, this PR was only tested manually.

Peek 2022-09-05 14-57

There are 12 events in this GIF:
- 3x `WHEEL_DOWN` (default behavior allowed = scrolling)
- 3x `Shift + WHEEL_DOWN` (default behavior suppressed = NO scrolling)
- 3x `WHEEL_DOWN` (default behavior allowed = scrolling)
- 3x `Shift + WHEEL_DOWN` (default behavior suppressed = NO scrolling)

This is the wheel test element that can be seen in the GIF:
```
private class WheelTestDiv extends DomDiv implements IDomWheelEventHandler {

	public WheelTestDiv() {

		appendText("wheel me");
		setStyle(CssStyle.BACKGROUND_COLOR, "yellow");

		CurrentDomDocument.get().getEngine().setPreventDefaultOnWheel(this, Set.of(DomModifier.CONTROL), true);
	}

	@Override
	public void handleWheel(IDomEvent event) {

		if (event.getDeltaY() > 0) {
			appendChild(new DomDiv()).appendText("down");
			Sleep.sleep(500);
		} else if (event.getDeltaY() < 0) {
			appendChild(new DomDiv()).appendText("up");
			Sleep.sleep(500);
		}
	}
}
```

### DIFF
--- a/platform-dom/src/main/java/com/softicar/platform/dom/engine/IDomEngine.java
+++ b/platform-dom/src/main/java/com/softicar/platform/dom/engine/IDomEngine.java
@@ -14,6 +14,7 @@ import com.softicar.platform.dom.elements.popup.DomPopup;
 import com.softicar.platform.dom.elements.popup.IDomPopupFrame;
 import com.softicar.platform.dom.event.DomEventType;
 import com.softicar.platform.dom.event.DomKeys;
+import com.softicar.platform.dom.event.DomModifier;
 import com.softicar.platform.dom.event.timeout.IDomTimeoutNode;
 import com.softicar.platform.dom.input.DomOption;
 import com.softicar.platform.dom.input.DomSelect;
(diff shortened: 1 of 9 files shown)


platform-dom/src/main/java/com/softicar/platform/dom/engine/IDomEngine.java
```diff
@@ -102,10 +104,60 @@ public interface IDomEngine {
 	 */
 	void setFireOnKeyUp(IDomNode node, DomEventType type, boolean enabled);
 
+	/**
+	 * Tells the client browser to not perform the default action if an event of
+	 * the given {@link DomEventType} occurs on the given {@link IDomNode}.
```
oliver-richers, line 109: I think this only works for key events, we should list which DomEventType instances that are.

oliver-richers (approved):

PR #483, opened 2022-09-07
PLAT-1119 Unify css resource and class files in Core Module

daniel-klose:


oliver-richers (approved):
consider adapting the title, since this only affects the core module

PR #489, opened 2022-09-08
PLAT-1152-2 Migrated V22 to V23

marcel-palmi:


oliver-richers (approved):
ugly as hell, should be removed ASAP again

PR #499, opened 2022-09-12
PLAT-1145 Added setRequired to IDomTextualInput

daniel-klose:
image


platform-dom/src/main/java/com/softicar/platform/dom/input/IDomTextualInput.java
```diff
@@ -91,6 +91,20 @@ public interface IDomTextualInput extends IDomFocusable, IDomValueInput<String>
...
 	IDomTextualInput setPlaceholder(IDisplayString placeholder);
 
+	// -------------------------------- required -------------------------------- //
+
+	/**
+	 * Manages the HTML required attribute.
+	 *
+	 * @param required
+	 *            whether the required attribute should be added or not (never
+	 *            <i>null</i>)
```
oliver-richers, line 101: primitive `boolean` can never be null

platform-dom/src/main/java/com/softicar/platform/dom/input/IDomTextualInput.java
```diff
@@ -91,6 +91,20 @@ public interface IDomTextualInput extends IDomFocusable, IDomValueInput<String>
...
+	 * @return this
+	 */
+	IDomTextualInput setRequired(boolean required);
+
+	boolean isRequired();
```
oliver-richers, line 106: should also have a Javadoc, see `isReadonly`

oliver-richers (changes requested):

platform-dom/src/main/java/com/softicar/platform/dom/input/IDomTextualInput.java
daniel-klose: well, I just adapted the read-only javadoc

platform-dom/src/main/java/com/softicar/platform/dom/input/IDomTextualInput.java
daniel-klose: see above

oliver-richers (approved):

PR #500, opened 2022-09-12
PLAT-1151 High-Level Input Customization Methods on EmfForeignEntityAttribute

oliver-richers:


platform-core-module/src/main/java/com/softicar/platform/core/module/permission/assignment/AGModuleInstancePermissionAssignmentTable.java
```diff
@@ -40,8 +40,7 @@ public void customizeAttributeProperties(IEmfAttributeList<AGModuleInstancePermi
 		attributes//
 			.editAttribute(AGModuleInstancePermissionAssignment.USER)
 			.setImmutable(true)
-			.setPredicateMandatory(EmfPredicates.always())
-			.setInputFactory(ModuleInstancePermissionAssignmentUserInput::new);
+			.setPredicateMandatory(EmfPredicates.always());
```
daniel-klose, line 43: Shouldnt there be a `.addFilter(user -> !user.isSystemUser());` here?

platform-workflow-module/src/main/java/com/softicar/platform/workflow/module/workflow/user/configuration/AGWorkflowUserConfigurationTable.java
```diff
@@ -51,8 +50,8 @@ public void customizeLoggers(EmfChangeLoggerSet<AGWorkflowUserConfiguration> log
 	public void customizeAttributeProperties(IEmfAttributeList<AGWorkflowUserConfiguration> attributes) {
 
 		attributes//
-			.editAttribute(AGWorkflowUserConfiguration.SUBSTITUTE)
-			.setInputFactory(UserInput::new);
```
daniel-klose, line 55: Can we remove the `UserInput` class?
It seems orphaned now, and a bit useless in general

daniel-klose (changes requested):

platform-workflow-module/src/main/java/com/softicar/platform/workflow/module/workflow/user/configuration/AGWorkflowUserConfigurationTable.java
oliver-richers: unfortunately, we cannot, it is use in EAS aside from EMF-attributes

platform-core-module/src/main/java/com/softicar/platform/core/module/permission/assignment/AGModuleInstancePermissionAssignmentTable.java
oliver-richers: hmm, not sure if really necessary, but I can do that

oliver-richers: ok, added, also added to `AGRoleUser` for symmetry

daniel-klose (approved):

platform-core-module/src/main/java/com/softicar/platform/core/module/permission/assignment/AGModuleInstancePermissionAssignmentTable.java
alexander-schmidt: I don't see the point of suppressing the system user only in two distinct places. What about the remaining 5000 user inputs, particularly in the payload software? IMO, we either need a systematic solution to this - or none.

oliver-richers: not sure if this has anything to do with this PR? I just retained the status quo and extended it to `AGRoleUser` for symmetry...so?

PR #508, opened 2022-09-13
PLAT-1164 Remove unused functions insertTextAtCaret and moveCaretToPosition

alexander-schmidt:


oliver-richers (approved):
I hope we really won't need this anymore. ;)

PR #510, opened 2022-09-13
PLAT-1156 Add several checks to DbTableValidator

alexander-schmidt:


platform-db-runtime/src/main/java/com/softicar/platform/db/runtime/table/validator/DbTableValidator.java
```diff
@@ -3,71 +3,145 @@
...
-		this.dataFields = this.table.getDataFields();
+		this.dataFields = table.getDataFields();
+		this.allKeys = table.getAllKeys();
 	}
 
 	public void validate() {
 
-		validateStringFields();
-		validateNullability();
+		var errors = new ErrorList();
```
oliver-richers, line 33: hmm, would be less code if me made `errors` a field...just saying

platform-db-runtime/src/main/java/com/softicar/platform/db/runtime/table/validator/DbTableValidator.java
```diff
@@ -3,71 +3,145 @@
...
+				errors
+					.add(//
 						"Field %s.`%s` defaults to NULL but it is not NULLABLE.",
 						table.getFullName(),
-						field.getTitle());
-				errors.add(message);
+						field.getName());
+			}
+
+			if (field.isNullable() && !field.hasDefault()) {
```
oliver-richers, line 77: what is the rationale that nullable fields need a default?

oliver-richers (changes requested):

platform-db-runtime/src/main/java/com/softicar/platform/db/runtime/table/validator/DbTableValidator.java
alexander-schmidt: How would I know? Quoting the MySQL docs at https://dev.mysql.com/doc/refman/8.0/en/data-type-defaults.html#data-type-defaults-implicit:
>  If the column can take NULL as a value, the column is defined with an explicit DEFAULT NULL clause.

oliver-richers: pfff, ok ;)

oliver-richers (approved):

PR #517, opened 2022-09-19
PLAT-1105 Eagerly refresh EmfDataTable

daniel-klose:


platform-emf/src/main/java/com/softicar/platform/emf/management/EmfManagementDiv.java
```diff
@@ -110,7 +111,7 @@ public IEmfDataTableDiv<R> getDataTableDiv() {
 	@Override
 	public void refresh(IDomRefreshBusEvent event) {
 
-		if (event.isAnyObjectChanged(entityTable.getValueClass())) {
+		if (event.isAnyObjectChanged(IEmfEntity.class)) {
```
oliver-richers, line 114: why `IEmfEntity.class`? we should rather check for nothing, i think

oliver-richers (changes requested):

oliver-richers (approved):
simple, nice

PR #543, opened 2022-10-11
PLAT-1182-1 Rework DomTimeInput

daniel-klose:
Finally, TimeInput is a single field and normalizes stuff
timeInput


platform-dom/src/main/java/com/softicar/platform/dom/elements/testing/node/tester/AbstractDomNodeTester.java
```diff
@@ -73,10 +73,7 @@ public AbstractDomNodeTester<N> setTimeInputValue(ITestMarker marker, String tim
 		if (elements.size() != 3) {
 			throw new IllegalArgumentException("Expected a time string in the form '<hours>:<minutes>:<seconds>' but got: %s".formatted(timeString));
 		}
-		findNode(marker)//
-			.setInputValue(DomTestMarker.TIME_INPUT_HOURS_INPUT, elements.get(0))
-			.setInputValue(DomTestMarker.TIME_INPUT_MINUTES_INPUT, elements.get(1))
-			.setInputValue(DomTestMarker.TIME_INPUT_SECONDS_INPUT, elements.get(2));
+		findNode(marker).setInputValue(DomTestMarker.TIME_INPUT, timeString);
```
oliver-richers, line 76: i think we should drop the parsing code above, otherwise we cannot test non-normalized input texts

platform-dom/src/main/java/com/softicar/platform/dom/elements/time/DomTimeInput.java
```diff
@@ -4,88 +4,98 @@
...
+		List<String> tokens = new ArrayList<>();
+		var inputText = timeInput.getValueOrNull();
+		if (inputText != null && !inputText.isEmpty()) {
+			tokens = new Tokenizer(':', '\\').tokenize(inputText);
+			if (tokens.size() > 3) {
+				throw new IllegalTimeSpecificationException(inputText);
+			}
+		}
+		return tokens;
+	}
```
oliver-richers, line 93: - we should not work with `getValueOrNull()` there is `getValueText()` which never returns `null`
- and we should use `isBlank()`
```suggestion
	private List<String> tokenizeInputValue() {
		if(timeInput.getValueText().isBlank()) {
			return Collections.emptyList();
		} else {
			var tokens = new Tokenizer(':', '\\').tokenize(timeInput.getValueText());
			if (tokens.size() > 3) {
				throw new IllegalTimeSpecificationException(inputText);
			}
			return tokens;
		}
	}
```

platform-dom/src/main/java/com/softicar/platform/dom/elements/time/DomTimeInput.java
```diff
@@ -4,88 +4,98 @@
...
+				while (tokens.size() < 3) {
+					tokens.add("0");
+				}
+				var hours = Integer.valueOf(tokens.get(0));
+				var minutes = Integer.valueOf(tokens.get(1));
+				var seconds = Integer.valueOf(tokens.get(2));
+				return Optional.of(new Time(hours, minutes, seconds));
 			} else {
-				throw new IllegalTimeSpecificationException(getValueAsString());
+				throw new IllegalTimeSpecificationException(timeInput.getValueOrNull());
```
oliver-richers, line 55:
```suggestion
				throw new IllegalTimeSpecificationException(timeInput.getValueText());
```

platform-dom/src/main/java/com/softicar/platform/dom/elements/time/DomTimeInput.java
```diff
@@ -4,88 +4,98 @@
...
 			}
 		} catch (Exception exception) {
-			throw new IllegalTimeSpecificationException(exception, getValueAsString());
+			throw new IllegalTimeSpecificationException(exception, timeInput.getValueOrNull());
```
oliver-richers, line 58:
```suggestion
			throw new IllegalTimeSpecificationException(exception, timeInput.getValueText());
```

platform-dom/src/main/java/com/softicar/platform/dom/elements/time/DomTimeInput.java
```diff
@@ -4,88 +4,98 @@
...
+		var tokens = tokenizeInputValue();
 		try {
-			Integer hours = hourInput.getValue().orElse(null);
-			Integer minutes = minuteInput.getValue().orElse(null);
-			Integer seconds = secondInput.getValue().orElse(null);
-
-			if (hours != null && minutes != null && seconds != null) {
-				return Optional.of(new Time(hours, minutes, seconds));
-			} else if (hours == null && minutes == null && seconds == null) {
+			if (tokens.isEmpty() || tokens.stream().allMatch(String::isEmpty)) {
```
oliver-richers, line 44: why `tokens.stream().allMatch(String::isEmpty)`? so the user can input `::` and it would be considered empty? IMO that should not be the case

platform-dom/src/main/java/com/softicar/platform/dom/elements/time/DomTimeInput.java
```diff
@@ -4,88 +4,98 @@
...
-
-		return "%s:%s:%s"
-			.formatted(//
-				hourInput.getTextualValue(),
-				minuteInput.getTextualValue(),
-				secondInput.getTextualValue());
-	}
-
-	private DomIntegerInput createInput(IDisplayString title, ITestMarker testMarker) {
+	private DomTextInput createInput(IDisplayString title, ITestMarker testMarker) {
```
oliver-richers, line 72: there is actually not need anymore for the parameters

oliver-richers (changes requested):

platform-dom/src/main/java/com/softicar/platform/dom/elements/time/DomTimeInput.java
daniel-klose: Good to know, wanted to replicate current behavior first

oliver-richers (approved):

PR #546, opened 2022-10-11
PLAT-1181-4 Re-style user menu and working-indicator

alexander-schmidt:
- Moved user menu to the right.
- Shrunk application-is-working indicator, and moved it to the top-right corner.
- Fixed broken fading.

Peek 2022-10-11 18-03


oliver-richers (dismissed):
very nice

PR #548, opened 2022-10-12
PLAT-1185 AjaxWaitingDivTest: Fix non-determinisms

alexander-schmidt:
- Assert waiting-indicator state with `FluentWait`.
- Do not test the thing with itself (i.e. avoid `waitForServer`).


oliver-richers (approved):
nice

PR #555, opened 2022-10-25
PLAT-1194 Check state of Maintenance in LoginDiv

daniel-klose:


platform-core-module/src/main/java/com/softicar/platform/core/module/page/service/login/PageServiceLoginDiv.java
```diff
@@ -182,6 +183,7 @@ private void checkIfMaintenanceIsConfiguredForToday() {
 					.where(AGMaintenanceWindow.EXPECTED_START.isGreater(today.getBegin()))
 					.where(AGMaintenanceWindow.EXPECTED_START.isLess(today.getEnd()))
 					.where(AGMaintenanceWindow.EXPECTED_END.isGreater(DayTime.now()))
+					.where(AGMaintenanceWindow.STATE.isEqual(AGMaintenanceStateEnum.PENDING.getRecord()))
```
alexander-schmidt, line 186: Would we now still display a message if a maintenance window is currently active?

alexander-schmidt (changes requested):

platform-core-module/src/main/java/com/softicar/platform/core/module/page/service/login/PageServiceLoginDiv.java
daniel-klose: yes, this is only for pending ones
THe code for active maintenances is some lines above

alexander-schmidt: okay

alexander-schmidt (approved):

PR #559, opened 2022-10-26
PLAT-1200 Retain URL Debugging Parameters

oliver-richers:
- this is about the parameters `debug`, `profiler`, `stacktrace` and `verbose`
- without this change, only the initial page request will respect the mentioned debugging parameters which makes it impossible to debug/profile clicks on buttons, etc.

### DIFF
--- a/platform-common/src/main/java/com/softicar/platform/common/network/url/Url.java
+++ b/platform-common/src/main/java/com/softicar/platform/common/network/url/Url.java
@@ -98,14 +98,21 @@ public class Url {
 
 	private static String convertToQueryString(Entry<String, List<String>> parameter) {
 
-		List<String> values = parameter.getValue();
-		if (values.isEmpty()) {
-			return UrlCoding.encodeUtf8(parameter.getKey());
+		var encodedKey = UrlCoding.encodeUtf8(parameter.getKey());
+		var values = parameter.getValue();
+		if (isEmpty(values)) {
+			return encodedKey;
 		} else {
 			return values//
 				.stream()
-				.map(value -> UrlCoding.encodeUtf8(parameter.getKey()) + "=" + UrlCoding.encodeUtf8(value))
+				.map(UrlCoding::encodeUtf8)
+				.map(encodedValue -> encodedKey + "=" + encodedValue)
 				.collect(Collectors.joining("&"));
 		}
+	}
+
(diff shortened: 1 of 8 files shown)


platform-common/src/main/java/com/softicar/platform/common/network/url/Url.java
```diff
@@ -98,14 +98,21 @@ public String getStartingFromPath() {
 
 	private static String convertToQueryString(Entry<String, List<String>> parameter) {
```
oliver-richers, line 99: changes to this method only aim to beatify parameters with empty value, e.g. instead of `&profiler=` only `&profiler` is added to the URL

platform-core-module/src/main/java/com/softicar/platform/core/module/page/PageUrlBuilder.java
```diff
@@ -1,36 +1,64 @@
...
 	public Url build() {
 
 		return new WebServiceUrlBuilder(PageServiceFactory.class)//
 			.addParameter(PageParameterParser.getPageParameter(), page.getAnnotatedUuid().toString())
 			.addParameter("moduleInstance", moduleInstance.getItemId().toString())
+			.addParameters(debuggingParameters)
 			.build();
 	}
+
+	private Map<String, String> extractDebuggingParameters() {
```
alexander-schmidt, line 46: The constants in here are already defined elsewhere.
E.g. `"profiler"` is defined in `EmfPageConnectionProfiler.PROFILER_PARAMETER`.
Each of those constants should only be defined in one single place.

alexander-schmidt (changes requested):
I think we should validate URL parameter retention in a Selenium test (e.g. in CoreModule). Should be fairly straightforward.

platform-core-module/src/main/java/com/softicar/platform/core/module/page/PageUrlBuilder.java
oliver-richers: no they cannot, because they belong to different subsystems

oliver-richers: updated

alexander-schmidt (approved):

PR #560, opened 2022-10-26
PLAT-1201 Explicit GC on Page Change

oliver-richers:
- this is a quick fix to prevent dangling refresh bus listeners to pile up, degrading performance
- we can consider to remove this quick fix later

### DIFF
--- a/platform-core-module/src/main/java/com/softicar/platform/core/module/page/PageHeaderAndContentDiv.java
+++ b/platform-core-module/src/main/java/com/softicar/platform/core/module/page/PageHeaderAndContentDiv.java
@@ -32,10 +32,10 @@ public class PageHeaderAndContentDiv extends DomDiv {
 
 		removeChildren();
 		CurrentDomPopupCompositor.get().closeAll();
-
 		CurrentDomDocument.get().getDeferredInitializationController().clear();
-
 		DbTableRowCaches.invalidateAll();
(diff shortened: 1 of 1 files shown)


platform-core-module/src/main/java/com/softicar/platform/core/module/page/PageHeaderAndContentDiv.java
```diff
@@ -32,10 +32,10 @@ public void openPage(PageNavigationEntry<?> linkEntry) {
 
 		removeChildren();
 		CurrentDomPopupCompositor.get().closeAll();
-
 		CurrentDomDocument.get().getDeferredInitializationController().clear();
-
 		DbTableRowCaches.invalidateAll();
+		System.gc();
```
alexander-schmidt, line 37: This might cut both ways: We could even introduce delays where there are currently none. Plus, it's non-deterministic.
But we can try this, for sure.

alexander-schmidt (approved):

platform-core-module/src/main/java/com/softicar/platform/core/module/page/PageHeaderAndContentDiv.java
oliver-richers: tested this locally, no difference

PR #562, opened 2022-10-26
PLAT-1196 DomRefreshBus: Defer refresh calls on disappended nodes

alexander-schmidt:


platform-dom/src/main/java/com/softicar/platform/dom/refresh/bus/DomRefreshBus.java
```diff
@@ -92,4 +90,35 @@ private void submitQueue() {
 		}
 		queue.clear();
 	}
+
+	private void convertDanglingNodeListeners() {
+
+		var bodyFinder = new DomParentNodeFinder<>(DomBody.class);
```
oliver-richers, line 96: - this is already the 2nd use of this check, IMO we should add something like `IDomNode.isAppendedToDocument()` and also use that in `DomDeferredInitializationController.java`
- please note that `isAppendedToDocument` is better than `isAppendedToBody`, because actually, the HTML standard defines a `document` element above the body element which we do not reflect in our tree because we don't need it

platform-dom/src/main/java/com/softicar/platform/dom/refresh/bus/DomRefreshBus.java
```diff
@@ -62,6 +59,7 @@ public IDomRefreshBus submitEvent() {
 
 		if (!refreshEvent.isEmpty()) {
 			submitQueue();
+			convertDanglingNodeListeners();
```
oliver-richers, line 62: just suggesting
```suggestion
			removeListenersNotAppendedToDocument();
```

platform-dom/src/main/java/com/softicar/platform/dom/refresh/bus/DomRefreshBus.java
```diff
@@ -92,4 +90,35 @@ private void submitQueue() {
...
+					.addDeferredInitializer(listener, new ReappendedNodeInitializer(listener));
+			}
+		}
+	}
+
+	private class ReappendedNodeInitializer implements INullaryVoidFunction {
+
+		private final IDomRefreshBusListener listener;
+
+		public ReappendedNodeInitializer(IDomRefreshBusListener listener) {
```
oliver-richers, line 112: the class name `ReappendedNodeInitializer` talks about a `node` while the parameter talks about a `listener`, IMO we should only use the term `listener` or `listeningNode` everywhere to avoid confusion

oliver-richers (changes requested):

platform-dom/src/main/java/com/softicar/platform/dom/node/initialization/DomDeferredInitializationController.java
```diff
@@ -50,34 +49,73 @@ public void addDeferredInitializer(IDomNode node, INullaryVoidFunction initializ
...
+		while (!queuedNodes.isEmpty()) {
+			new AppendedNodeInitializer().initialize(flushQueuedNodes());
 		}
+	}
 
-		Optional//
-			.ofNullable(initializers.remove(document.getBody()))
-			.orElse(Collections.emptyList())
-			.forEach(INullaryVoidFunction::apply);
+	private ArrayList<IDomNode> flushQueuedNodes() {
```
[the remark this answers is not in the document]
alexander-schmidt: okay (refactoing artifact)

platform-dom/src/main/java/com/softicar/platform/dom/refresh/bus/DomRefreshBus.java
alexander-schmidt: I noticed that we already have `IDomNode.isAppended`, with an implementation in `AbstractDomNode` (which relies on `DomBody.isAppended` and `DomHead.isAppended` always returning `true`).
`DomRefreshBus` and `DomDeferredInitializationController` now use `IDomNode.isAppended`.

platform-dom/src/main/java/com/softicar/platform/dom/refresh/bus/DomRefreshBus.java
alexander-schmidt: Went with `convertListenersNotAppendedToDocument`.

platform-dom/src/main/java/com/softicar/platform/dom/refresh/bus/DomRefreshBus.java
alexander-schmidt: granted

platform-dom/src/main/java/com/softicar/platform/dom/refresh/bus/DomRefreshBusEvent.java
```diff
@@ -61,19 +61,22 @@ public boolean isAllChanged() {
 		return allChanged;
 	}
```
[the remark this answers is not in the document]
alexander-schmidt: Wow. Missed that. Thanks.

oliver-richers (approved):

PR #569, opened 2022-10-27
PLAT-1182-2 Rework temporal input elements

alexander-schmidt:
- `DomDayTimeInput` is now realized as a single text input field.
- `DomDayInput`, `DomDayTimeInput` and `DomTimeInput` now normalize their values on change.
- Merged various input-related CSS definitions into `dom-value-input-style.css`.
- Extracted and reworked input parsing code from `DomTimeInput` and `DomDayTimeInput`.
  - Wrote `DomTimeInputParser[Test]` and `DomDayTimeInputParser[Test]`.
  - Deliberately retained `DayTimeParser` (which is far more strict and also used in non-UI logic) as a separate implementation.
  - We could consider merging `DomDayTimeInputParser` and `DayTimeParser` later-on; but that's too much of a risk in the scope of this PR.
- Created `DomValueInputTooltip`.
  - An instantly-appearing pseudo tooltip, meant to be used on textual input elements.
  - Added format-hinting pseudo tooltips to `DomDayInput`, `DomDayTimeInput` and `DomTimeInput`.
- Added support for localized date and time formats in `ILocale` and `AGLocalization`.
  - Otherwise we would, for example, display `dd.MM.yyyy` to a German user. Displaying `TT.MM.JJJJ` to them is more appropriate.

Peek 2022-10-27 15-01


daniel-klose (approved):

platform-common/src/main/java/com/softicar/platform/common/core/locale/ILocale.java
```diff
@@ -49,4 +49,35 @@ public interface ILocale {
...
+	/**
+	 * Returns a human-readable time format that uses abbreviations which
+	 * correspond to the language returned by {@link #getLanguage()}.
+	 *
+	 * @return a localized, human-readable time format (never <i>null</i>)
+	 */
+	String getLocalizedTimeFormat();
+
+	/**
+	 * Returns a human-readable daytime format that uses abbreviations which
```
oliver-richers, line 70: i think the term `daytime` is not correct here either refer to `{@link DayTime}` or talk about `date and time format` or `format for a point in time`, etc

platform-common/src/main/java/com/softicar/platform/common/core/locale/ILocale.java
```diff
@@ -49,4 +49,35 @@ public interface ILocale {
...
+	 * correspond to the language returned by {@link #getLanguage()}.
+	 * <p>
+	 * By default, a space-separated concatenation of
+	 * {@link #getLocalizedDateFormat()} and {@link #getLocalizedTimeFormat()}
+	 * is returned.
+	 *
+	 * @return a localized, human-readable daytime format (never <i>null</i>)
```
oliver-richers, line 77: again `daytime`

platform-common/src/main/java/com/softicar/platform/common/date/CommonDateI18n.java
```diff
@@ -20,12 +20,10 @@ public interface CommonDateI18n extends CommonMonthI18n, CommonWeekdayI18n {
 		.de("%smin %ss");
 	I18n1 ILLEGAL_DATE_SPECIFICATION_ARG1 = new I18n1("Illegal date specification: %s")//
 		.de("Ungültige Datumsangabe: %s");
+	I18n1 ILLEGAL_DAYTIME_SPECIFICATION_ARG1 = new I18n1("Illegal daytime specification: %s")//
```
oliver-richers, line 23: again the term `daytime`

platform-common/src/main/java/com/softicar/platform/common/date/IllegalDayTimeSpecificationException.java
```diff
@@ -0,0 +1,21 @@
+package com.softicar.platform.common.date;
+
+import com.softicar.platform.common.core.exceptions.SofticarUserException;
+
+/**
+ * Thrown if an illegal daytime value was specified.
```
oliver-richers, line 6: again `daytime` -> `{@link DayTime}`

platform-common/src/main/java/com/softicar/platform/common/core/locale/Locale.java
```diff
@@ -13,13 +14,17 @@ public class Locale implements ILocale {
 	private String decimalSeparator;
 	private String digitGroupSeparator;
 	private String dateFormat;
+	private String localizedDateFormat;
+	private String localizedTimeFormat;
```
oliver-richers, line 18: - why do we have `dateFormat` and `localizedDateFormat`?
- not sure why we need the `localized` prefix? everything in here is localized

platform-common/src/main/java/com/softicar/platform/common/core/locale/Locale.java
```diff
@@ -13,13 +14,17 @@ public class Locale implements ILocale {
...
 
 	public Locale() {
 
 		this.language = LanguageEnum.ENGLISH;
 		this.decimalSeparator = ".";
 		this.digitGroupSeparator = "";
 		this.dateFormat = "yyyy-MM-dd";
+		this.localizedDateFormat = "YYYY-MM-DD";
+		this.localizedTimeFormat = "HH:MM:SS";
```
oliver-richers, line 27: I am confused.
```suggestion
		this.dateFormat = "yyyy-MM-dd";
		this.timeFormat = "HH:MM:SS";
```

platform-core-module/src/main/java/com/softicar/platform/core/module/localization/AGLocalization.sqml
```diff
@@ -9,6 +9,8 @@ TABLE AGLocalization 'Core.Localization' {
 	String decimalSeparator [MAXLENGTH=255]
 	String digitGroupSeparator = '' [MAXLENGTH=255]
 	String dateFormat [MAXLENGTH=255]
+	String localizedDateFormat [MAXLENGTH=255]
+	String localizedTimeFormat [MAXLENGTH=255]
```
oliver-richers, line 13: again, confused
```suggestion
	String dateFormat [MAXLENGTH=255]
	String timeFormat [MAXLENGTH=255]
```

oliver-richers (changes requested):
not finishing review until some stuff is cleared

platform-common/src/main/java/com/softicar/platform/common/core/locale/Locale.java
alexander-schmidt: `dateFormat` contains format codes of `java.text.SimpleDateFormat` which will appear odd or even incomprehensible to ordinary mortals; particularly if they use the system in a non-English language:
- When `yyyy-MM-dd` is displayed to an English-speaking user, they will probably wonder why it's neither `yyyy-mm-dd` nor `YYYY-MM-DD`. It looks inconsistent, and like a mistake.
- When `dd.MM.yyyy` is displayed to a German-speaking user, capitalization might be the least of their issues because they'd expect something like `TT.MM.JJJJ`. They might even fail to comprehend the format because of this; so displaying it (which we do in those pseudo tooltips) would be kind of pointless.

That being said, quoting from the Javadoc of `ILocale`:
```
/**
 * Returns a human-readable date format that uses abbreviations which
 * correspond to the language returned by {@link #getLanguage()}.
 *
 * @return a localized, human-readable date format (never <i>null</i>)
 */
String getLocalizedDateFormat();

/**
 * Returns a human-readable time format that uses abbreviations which
 * correspond to the language returned by {@link #getLanguage()}.
 *
 * @return a localized, human-readable time format (never <i>null</i>)
 */
String getLocalizedTimeFormat();
```

I'm not entirely happy with the naming either... I used "informal" instead of "localized" before. Any thoughts on that?

PR #571, opened 2022-10-28
PLAT-1203 Image Converter for Formats Not Supported by Browser

oliver-richers:


platform-ajax/src/test/java/com/softicar/platform/ajax/engine/AjaxDomEngineTest.java
```diff
@@ -11,10 +15,37 @@
...
+	 */
+	@Test
+	public void testSetAlternateResourceOnErrorWithTiff() {
+
+		var tiff = openTestNode(() -> new DomImage(AjaxTestResources.TEST_TIFF.getResource()));
+
+		// TIFF converted to PNG
+		var tiffResource = assertOne(getResourceById(getAttributeValue(tiff, "src")));
+		assertEquals(MimeType.IMAGE_PNG, tiffResource.getMimeType());
+		assertEquals("test.tiff.png", assertOne(tiffResource.getFilename()));
```
alexander-schmidt, line 46: Shouldn't we assert the content of the output image - somewhere, and at least heuristically?
E.g. we could assert the dimensions of the output and the color of the pixel at `(0,0)`. That should be fairly straightforward, and it would be sufficient to ensure that we don't end up with an empty 1x1 image, with a black box, or something like this. `DomImageResourceConverterTest` would be an appropriate place for that.

platform-ajax/src/main/java/com/softicar/platform/ajax/resource/AjaxResourceUrlParser.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.regex.Pattern;
+
+/**
+ * Extracts resource hash and resource ID from resource URLs.
+ *
+ * @author Oliver Richers
+ */
+public class AjaxResourceUrlParser {
+
+	private static final Pattern HASH_PATTERN = Pattern.compile(".*\\?resourceHash=([0-9a-fA-F]+).*");
```
alexander-schmidt, line 14: - Should we support URLs in which `resourceHash` is _not_ the first parameter? Like so: `\\?` --> `[\\?&]`
- I think we should reuse the constant `AjaxResourceUrl.HASH_PARAMETER_NAME` here.

platform-ajax/src/main/java/com/softicar/platform/ajax/resource/AjaxResourceUrlParser.java
```diff
@@ -0,0 +1,42 @@
...
+	private static final Pattern ID_PATTERN = Pattern.compile(".*\\?resourceId=([0-9]+).*");
```
alexander-schmidt, line 15: See above.

platform-ajax/src/main/java/com/softicar/platform/ajax/engine/AjaxDomEngine.java
```diff
@@ -434,6 +436,14 @@ public void setWorkingIndicatorEnabled(boolean enabled) {
 		JS_call("setWorkingIndicatorEnabled", enabled);
 	}
 
+	@Override
+	public void setAlternateResourceOnError(DomImage image, IResource resource) {
+
+		Objects.requireNonNull(image);
+		Objects.requireNonNull(resource);
+		addJavascript("%s.onerror=function(){%s.src=%s;%s.onerror=null;};", image, image, getResourceUrl(resource), image);
```
alexander-schmidt, line 444: This approach, while certainly functional (pun intended), appears uncommon. Shouldn't we add a feature to the TypeScript code/API, instead of injecting JS code?

alexander-schmidt (changes requested):

platform-ajax/src/main/java/com/softicar/platform/ajax/resource/AjaxResourceUrlParser.java
oliver-richers: did so

platform-ajax/src/main/java/com/softicar/platform/ajax/resource/AjaxResourceUrlParser.java
oliver-richers: did so, even though looks ugly now ;)

platform-ajax/src/main/java/com/softicar/platform/ajax/engine/AjaxDomEngine.java
oliver-richers: I though about it, but in the end it would only make things more complicated, IMO

alexander-schmidt: Well, we could as well change this later-on. Okay for now.

platform-ajax/src/test/java/com/softicar/platform/ajax/engine/AjaxDomEngineTest.java
oliver-richers: did so

alexander-schmidt (approved):

PR #573, opened 2022-11-01
PLAT-1197 DomRefreshBus and New DbTableRowNotificationType.CHANGE

oliver-richers:


platform-db/src/main/java/com/softicar/platform/db/runtime/table/listener/DbTableRowNotificationType.java
```diff
@@ -10,16 +10,16 @@
 public enum DbTableRowNotificationType {
 
 	/**
-	 * Denotes that a {@link IDbTableRow} was deleted from the database, i.e.
-	 * <i>DELETE</i>.
+	 * Denotes that a {@link IDbTableRow} was persistently changed, that is, it
+	 * was saved with changes or loaded with changes or just deleted.
```
alexander-schmidt, line 14:
```suggestion
	 * Denotes that a {@link IDbTableRow} was persistently changed. That is, it
	 * was saved with changes, or loaded with changes, or deleted.
```

platform-db/src/main/java/com/softicar/platform/db/runtime/table/logic/save/DbTableRowSaverSlave.java
```diff
@@ -51,6 +53,17 @@ public void afterSave() {
 			sendNotification(IDbTableListener::afterSave, rows);
 			DbTableRowCommitNotifier.addNotification(table, DbTableRowNotificationType.SAVE, rows);
 		}
+		if (!changedRows.isEmpty()) {
+			DbTableRowCommitNotifier.addNotification(table, DbTableRowNotificationType.CHANGE, changedRows);
+		}
+	}
+
+	private Collection<R> getRowsWithPersistentChange(Collection<R> rows) {
```
alexander-schmidt, line 61: This mehtod name got me confused. How about plain and simple `getChangedRows`, or `determineChangedRows`?

platform-db/src/main/java/com/softicar/platform/db/runtime/table/logic/DbTableRowDeleter.java
```diff
@@ -84,6 +84,7 @@ private void deleteRows() {
 
 			// send post-delete notification
 			sendNotification(IDbTableListener::afterDelete, rows);
+			DbTableRowCommitNotifier.addNotification(table, DbTableRowNotificationType.CHANGE, rows);
```
alexander-schmidt, line 87: The `[I]DbTableRowNotificationSet` can now (obviously) contain several differently-typed notifications for any given row. I wonder if we should warn about that in the Javadoc of `IDbTableRowNotificationSet.[getDeletedRows|getChangedRows]` because calling code might need to implement deduplication.

alexander-schmidt (changes requested):

platform-db/src/main/java/com/softicar/platform/db/runtime/table/listener/DbTableRowNotificationType.java
oliver-richers: - added the commas but not the point (i think `that is` is usually appended by a comma)
- retained the `just` to express, that the delete does not require the row to be changed, as for saving and loading

platform-db/src/main/java/com/softicar/platform/db/runtime/table/logic/save/DbTableRowSaverSlave.java
oliver-richers: - it is not just `ChangedRows` because it does not only look for `row.dataChanged()` but also `row.impermanent()`
- open for suggestions

platform-db/src/main/java/com/softicar/platform/db/runtime/table/logic/DbTableRowDeleter.java
oliver-richers: added Javadoc to `IDbTableRowNotificationSet`

platform-db/src/main/java/com/softicar/platform/db/runtime/table/listener/DbTableRowNotificationType.java
alexander-schmidt: okay

platform-db/src/main/java/com/softicar/platform/db/runtime/table/logic/DbTableRowDeleter.java
alexander-schmidt: okay

platform-db/src/main/java/com/softicar/platform/db/runtime/table/logic/save/DbTableRowSaverSlave.java
alexander-schmidt: We kind of overloaded the semantics of "change" here: On the one hand we have `DbTableRowFlag.DATA_CHANGED`, and on the other hand we have `DbTableRowNotificationType.CHANGE`.

We could resolve this in a brutally explicit way:
`this.rowsForChangeNotification = determineRowsForChangeNotification(rows);`

Not quite elegant but at least it leaves no room for interpretation.

oliver-richers: good idea, changed as requested

alexander-schmidt (approved):

PR #581, opened 2022-11-09
PLAT-1213 Eagerly refresh invalidated or stubbed table rows

alexander-schmidt:


platform-db/src/main/java/com/softicar/platform/db/runtime/table/logic/AbstractDbTableRowFieldAccessor.java
```diff
@@ -29,7 +30,7 @@ private void reloadRowIfNecessary() {
 
 		if (!row.table().getPrimaryKey().containsField(field) && !row.impermanent()) {
 			if (row.stub() || row.invalidated()) {
-				row.reload();
+				row.table().refreshAll(Collections.singleton(row));
```
daniel-klose, line 33: Hm, is this a common occurence that we refresh a single row?
If it is, maybe we should add a default implementation of `refreshAll` with a varArgs-Param

platform-db/src/main/java/com/softicar/platform/db/runtime/cache/DbTableRowCacheBurstSizeCalculator.java
```diff
@@ -0,0 +1,86 @@
...
+		return this;
+	}
+
+	public DbTableRowCacheBurstSizeCalculator setBurstSizeCooldown(Duration burstSizeCooldown) {
+
+		this.burstSizeCooldown = burstSizeCooldown;
+		return this;
+	}
+
+	public int getNextBurstSize() {
```
oliver-richers, line 56: i think that the name `get` rather implies that the method has no side-effect, so I suggest to rename it
```suggestion
	public int calculateNextBurstSize() {
```

oliver-richers (approved):

platform-db/src/main/java/com/softicar/platform/db/runtime/table/logic/AbstractDbTableRowFieldAccessor.java
alexander-schmidt: As this is rather uncommon (beyond unit tests), I don't really see a point in adding yet another method to `IDbBasicTable`.

platform-db/src/main/java/com/softicar/platform/db/runtime/cache/DbTableRowCacheBurstSizeCalculator.java
alexander-schmidt: ok

PR #584, opened 2022-11-12
PLAT-1216 Image to PDF Converter

oliver-richers:


Jerome-Feininger (dismissed):

platform-common/src/main/java/com/softicar/platform/common/ui/image/Images.java
```diff
@@ -0,0 +1,69 @@
...
+import java.awt.image.BufferedImage;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Supplier;
+import javax.imageio.ImageIO;
+
+/**
+ * Convenience methods for {@link ImageIO}.
```
alexander-schmidt, line 14: - If this class is indeed ImageIO specific - should we reflect that in the class name? e.g. `ImageIOImages` or so
- Either way, I think we should explicitly mention that supported formats depend on what's supported by `ImageIO`.

platform-common/src/main/java/com/softicar/platform/common/ui/image/Images.java
```diff
@@ -0,0 +1,69 @@
...
+	 *
+	 * @param inputStreamFactory
+	 *            the factory for the {@link InputStream} (never <i>null</i>)
+	 * @return the {@link List} of loaded {@link BufferedImage} objects
+	 */
+	public static List<BufferedImage> readImages(Supplier<InputStream> inputStreamFactory) {
+
+		var images = new ArrayList<BufferedImage>();
+		try (var inputStream = inputStreamFactory.get()) {
+			try (var imageInputStream = ImageIO.createImageInputStream(inputStream)) {
```
alexander-schmidt, line 53: We could save an indentation level by merging the try-block headers.

platform-common/src/main/java/com/softicar/platform/common/ui/image/Images.java
```diff
@@ -0,0 +1,69 @@
...
+				var readers = ImageIO.getImageReaders(imageInputStream);
+				if (readers.hasNext()) {
+					var reader = readers.next();
+					reader.setInput(imageInputStream);
+					var count = reader.getNumImages(true);
```
alexander-schmidt, line 58: I think we should inline `count`.

alexander-schmidt (changes requested):

platform-common-pdf/src/main/java/com/softicar/platform/common/pdf/ImageToPdfConverter.java
```diff
@@ -0,0 +1,72 @@
...
+	public ImageToPdfConverter(Supplier<InputStream> inputStreamFactory) {
+
+		this.inputStreamFactory = Objects.requireNonNull(inputStreamFactory);
+	}
+
+	/**
+	 * Adds the image(s) to the given {@link PDDocument}.
+	 *
+	 * @param document
+	 *            the {@link PDDocument} (never {@code null})
```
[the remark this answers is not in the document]
oliver-richers: changed to `<i>null</i>`

platform-common-pdf/src/main/java/com/softicar/platform/common/pdf/ImageToPdfConverter.java
```diff
@@ -0,0 +1,72 @@
...
+	 */
+	public void addToPdf(PDDocument document) {
```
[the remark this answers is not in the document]
oliver-richers: I thought it could be useful, but to avoid unnecessary discussions, changed `addToPdf` to `private`

platform-common/src/main/java/com/softicar/platform/common/ui/image/Images.java
oliver-richers: `ImageIO` is an integral part of the Java API, IMO, everybody working with images in Java will use it by default.

platform-common/src/main/java/com/softicar/platform/common/ui/image/Images.java
oliver-richers: IMO, it is easier to read with one resource per try-block

platform-common/src/main/java/com/softicar/platform/common/ui/image/Images.java
alexander-schmidt: Okay

platform-common/src/main/java/com/softicar/platform/common/ui/image/Images.java
alexander-schmidt: okay

alexander-schmidt (approved):

PR #586, opened 2022-11-14
PLAT-1218 Fixed TheadLocal Leaks and Other Improvements

oliver-richers:


platform-core-module/src/main/java/com/softicar/platform/core/module/ajax/listener/BatikCleanerThreadManager.java
```diff
@@ -11,25 +12,28 @@ public class BatikCleanerThreadManager {
...
 			threadField.setAccessible(true);
 			Thread thread = (Thread) threadField.get(null);
 			if (thread != null) {
 				if (new ThreadKiller<>(thread).setTimeout(5000).killAll()) {
-					Log.finfo("successfull shutdown of %s", BATIK_CLEANER_THREAD_CLASS);
+					Log.finfo("Successfull shutdown of %s", BATIK_CLEANER_THREAD_CLASS);
 				} else {
-					Log.finfo("failed shutdown of %s (stupid thread does not want to die)", BATIK_CLEANER_THREAD_CLASS);
-					Log.finfo("calling Thread.stop()");
+					Log.finfo("Failed shutdown of %s (stupid thread does not want to die)", BATIK_CLEANER_THREAD_CLASS);
```
alexander-schmidt, line 25: I think we should reword `stupid thread does not want to die`

platform-common/src/main/java/com/softicar/platform/common/core/singleton/CurrentSingletonSet.java
```diff
@@ -37,4 +37,12 @@ public static void set(SingletonSet singletonSet) {
 
 		INSTANCE.set(singletonSet);
 	}
+
+	/**
+	 * Removes the current {@link SingletonSet}.
+	 */
+	public static void reset() {
```
alexander-schmidt, line 44: So, shouldn't this method be called `remove`, or `clear`?

platform-core-module/src/main/java/com/softicar/platform/core/module/ajax/listener/AjaxContextListener.java
```diff
@@ -47,5 +48,6 @@ public void contextDestroyed(ServletContextEvent sce) {
 		new IioRegistryManager().deregisterProviders();
 		servletContext = null;
 		Log.finfo("CONTEXT DESTROYED: %s", sce.getServletContext().getContextPath());
+		CurrentSingletonSet.reset();
```
alexander-schmidt, line 51: Seems like we want to guarantee this call to happen. If so, we should probably wrap everything above in a `try` block, and call `CurrentSingletonSet.reset();` in a corresponding `finally` block.

alexander-schmidt (changes requested):

platform-core-module/src/main/java/com/softicar/platform/core/module/ajax/listener/BatikCleanerThreadManager.java
oliver-richers: ok, removed

platform-common/src/main/java/com/softicar/platform/common/core/singleton/CurrentSingletonSet.java
oliver-richers: renamed to `remove`

platform-core-module/src/main/java/com/softicar/platform/core/module/ajax/listener/AjaxContextListener.java
oliver-richers: added `try...finally`

alexander-schmidt (approved):

PR #597, opened 2022-11-24
PLAT-1230 Use H2 databases in case-insensitive mode

alexander-schmidt:


platform-db/src/main/java/com/softicar/platform/db/core/dbms/h2/DbH2MemoryConnector.java
```diff
@@ -43,6 +43,7 @@ protected Properties createDefaultProperties(IDbDatabase database) {
 		Properties properties = new Properties();
 		properties.put("DATABASE_TO_UPPER", databaseToUpper? "true" : "false");
 		properties.put("MODE", "MySQL");
+		properties.put("IGNORECASE", "TRUE");
```
oliver-richers, line 46: did you also execute all tests in EAS?

oliver-richers (approved):

platform-db/src/main/java/com/softicar/platform/db/core/dbms/h2/DbH2MemoryConnector.java
alexander-schmidt: Sure. All green, as expected.

PR #605, opened 2022-12-07
PLAT-1238-4 DomAutoCompleteInput: Reuse the cache for value and state in deduceValue()

alexander-schmidt:
By no longer calling the expensive `DomAutoCompleteValueParser.parse` method but instead reusing the value-and-state cache in `DomAutoCompleteInput.deduceValue`, the following effects were achieved in a real-world test case, while processing a single DOM event:
- The total time consumption across 16 combined calls to `deduceValue()` and `getValue()` could be reduced from `1050ms` to `534ms` (i.e. cut by `49%`).
- Only one of those 16 combined calls resulted in a `DomAutoCompleteValueParser.parse` call.


oliver-richers (dismissed):

oliver-richers (approved):
better

PR #635, opened 2023-01-05
EXP-2382 Add customizeFormAttributesDiv to IEmfTable

viktor-nikolaus:


platform-emf/src/main/java/com/softicar/platform/emf/table/IEmfTable.java
```diff
@@ -367,6 +368,11 @@ default void customizeEmfTableConfiguration(EmfTableConfiguration<R, P, S> confi
 		DevNull.swallow(configuration);
 	}
 
+	default void customizeFormAttributesDiv(EmfFormAttributesDivConfiguration<R> attributesDivConfiguration) {
+
+		DevNull.swallow(attributesDivConfiguration);
+	}
+
```
viktor-nikolaus, line 375: Not sure if extending `EmfTableConfiguration` would have been better.

alexander-schmidt: I think so. This way we could get rid of `[I]EmfFormAttributesDivConfiguration` which is oddly similar to `[I]EmfFormAttributesDivFactory` anyway.

alexander-schmidt (changes requested):

alexander-schmidt (approved):

PR #647, opened 2023-01-11
PLAT-1268 Optimized Test Fixtures

oliver-richers:
Optimizations:
- burst insertion for `ProgramStandardConfiguration`
- burst insertion for `WorkflowIconStandardConfiguration`
- `TestFixtureRegistry` now disables EMF logging through `EmfTableListenerSettings`
- `TestFixtureRegistry` now wraps test fixtures in transations
- implemented new `UuidCache`
- `DbAutomaticTableCreator` now executes way less `CREATE SCHEMA` statements
- disabled table listener for `AGUuid`


platform-core-module/src/main/java/com/softicar/platform/core/module/uuid/UuidCache.java
```diff
@@ -0,0 +1,58 @@
+package com.softicar.platform.core.module.uuid;
+
+import com.softicar.platform.db.core.connection.DbConnectionOverrideScope;
+import java.util.Map;
+import java.util.UUID;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+class UuidCache {
```
alexander-schmidt, line 9: We should have a test for this, since it's an extremely important core mechanism.

platform-emf/src/main/java/com/softicar/platform/emf/table/listener/EmfTableListenerSettings.java
```diff
@@ -0,0 +1,29 @@
+package com.softicar.platform.emf.table.listener;
+
+import com.softicar.platform.db.core.connection.DbConnections;
+
+public class EmfTableListenerSettings {
```
alexander-schmidt, line 5: - I think this deserves some Javadoc since it relies on the rather unusual principle of interacting with the current transaction (which is not obvious at all).
- I think we should have a test for this.

platform-db/src/main/java/com/softicar/platform/db/runtime/table/creator/DbAutomaticTableCreator.java
```diff
@@ -148,4 +156,15 @@ private IDbConnection getConnection() {
 		}
 		return connection;
 	}
+
+	private void createDatabaseIfMissing(IDbTable<?, ?> table) {
+
+		var databaseName = table.getFullName().getDatabaseName();
+		if (!existingDatabases.contains(databaseName.toLowerCase())) {
+			new DbStatement()//
+				.addText(String.format("CREATE SCHEMA IF NOT EXISTS `%s`", databaseName))
```
alexander-schmidt, line 165: I assume that the major performance benefit here comes from reducing the number of times this statement is executed. Is that correct?

platform-core-module/src/main/java/com/softicar/platform/core/module/standard/configuration/ProgramStandardConfiguration.java
```diff
@@ -1,50 +1,72 @@
...
 	@Override
 	public void createAndSaveAll() {
 
-		Programs//
+		var programs = createPrograms();
+
+		saveAll(programs);
+		saveAll(createStates(programs));
+		saveAll(createScheduledExecutions(programs));
+	}
```
alexander-schmidt, line 24: Out of curiosity: How much time did we save here?

alexander-schmidt (changes requested):

platform-db/src/main/java/com/softicar/platform/db/runtime/table/creator/DbAutomaticTableCreator.java
oliver-richers: yes, that is the primary aim of this PR

platform-core-module/src/main/java/com/softicar/platform/core/module/standard/configuration/ProgramStandardConfiguration.java
oliver-richers: I did not measure this in isolation, I rather looked at the amount of statements, and this were a lot of statements, including unnecessary BEGIN, COMMIT and SELECT statements.

platform-emf/src/main/java/com/softicar/platform/emf/object/table/EmfObjectTable.java
```diff
@@ -20,12 +20,19 @@
...
 		super(builder);
 
 		this.configuration = new EmfTableConfiguration<>(this);
-
-		addTableListener(new EmfTableListener<>(this));
 		customizeEmfTableConfiguration(configuration);
+
+		if (addTableListener) {
+			addTableListener(new EmfTableListener<>(this));
+		}
```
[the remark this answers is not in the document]
oliver-richers: If we would spend the amount to implement such a check and warning, we could instead only add the table listener whenever someone adds a hook, but such a logic is too costly to implement for now. I would rather get rid of `AGUuid` and this flag. `AGUuid` is a very special table, and this flag should only be used there.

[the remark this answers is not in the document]
alexander-schmidt: nvm, not worth the effort

platform-core-module/src/main/java/com/softicar/platform/core/module/uuid/UuidCache.java
oliver-richers: implemented

platform-emf/src/main/java/com/softicar/platform/emf/table/listener/EmfTableListenerSettings.java
oliver-richers: added javadoc and test

alexander-schmidt (approved):

PR #657, opened 2023-01-16
PLAT-1277 Fix calmping and scrolling with percent based popup placement

alexander-schmidt:
**Behavior with this PR:**
Peek 2023-01-16 11-29


oliver-richers (approved):
nice

PR #670, opened 2023-01-23
PLAT-1279 EmailToPdfConverter: Implement conversion of MSG files

alexander-schmidt:
- Implemented MSG-to-PDF conversion by using EML as an intermediate format.
- Added `simpleJavaMail` dependency and upgraded others as necessary.
  - Particularly, `jakarta[Actication|Mail]` had to be upgraded to `2.x`.
- Added `StoredFileConverter`.
- Added unit tests.


platform-core-module/src/main/java/com/softicar/platform/core/module/file/stored/StoredFileConverter.java
```diff
@@ -0,0 +1,44 @@
...
+import com.softicar.platform.core.module.email.converter.EmailToPdfConverter;
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * Facilitates conversion of {@link AGStoredFile} instances.
+ *
+ * @author Alexander Schmidt
+ */
+public class StoredFileConverter {
```
oliver-richers, line 13: hmm, this API makes it hard to find it...I suggest to make this class package-private and add a `AGStoredFile.toPdf` method that we can then extend to support more and more formats in the future.

oliver-richers (approved):

platform-core-module/src/main/java/com/softicar/platform/core/module/file/stored/StoredFileConverter.java
alexander-schmidt: Did something similar. See `StoredFileEmailPreviewPopup` as an example:
`file.convert().fromEmailToPdfBytes().ifPresentOrElse(this::appendPdfDisplay, this::appendFailureMessage);`

oliver-richers (approved):
nice

PR #675, opened 2023-01-26
PLAT-1290-1 StoredFileConverter: Generalize semantics and add isConvertibleToPdf()

alexander-schmidt:


platform-core-module/src/main/java/com/softicar/platform/core/module/file/stored/StoredFileConverter.java
```diff
@@ -22,36 +20,61 @@ public class StoredFileConverter {
...
-	 * If no conversion strategy can be identified for the {@link AGStoredFile},
-	 * {@link Optional#empty()} is returned.
+	 * Supports the following {@link AGStoredFile} formats:
+	 * <ul>
+	 * <li>EML</li>
+	 * <li>MSG</li>
+	 * </ul>
+	 * If the {@link AGStoredFile} format is not recognized,
+	 * {@link Optional#empty()} is returned. In that case,
+	 * {@link #isConvertibleToPdf()} is guaranteed to return <i>false</i>.
```
oliver-richers, line 38:
```suggestion
	 * If the {@link AGStoredFile} format is not recognized,
	 * {@link Optional#empty()} is returned. In that case,
	 * {@link #isConvertibleToPdf()} is guaranteed to return <i>false</i>.
```
```suggestion
	 * If and only if {@link #isConvertibleToPdf()} returns <i>false</i>,
	 * the {@link AGStoredFile} cannot be converted to PDF and {@link Optional#empty()}
	 * is returned. 
```

oliver-richers (approved):

oliver-richers (approved):

PR #693, opened 2023-02-03
PLAT-1302 TesseractOcrTextExtractor: Avoid PNG codec overhead

alexander-schmidt:
Use BMP to exchange data between the PDF renderer and Tesseract, in order to avoid PNG codec overhead.
Saves around 200 to 500ms per page. Performance gains depend on the input file.


oliver-richers (approved):
nice

PR #694, opened 2023-02-03
PLAT-1303 AGStoredFile: Facilitate setting the content type by IMimeType

alexander-schmidt:


platform-core-module/src/main/java/com/softicar/platform/core/module/file/stored/AGStoredFile.java
```diff
@@ -210,6 +210,19 @@ public boolean hasMimeTypeOrExtension(MimeType mimeType) {
...
 	}
 
+	/**
+	 * Sets the given {@link IMimeType} as {@link AGStoredFile#CONTENT_TYPE}.
+	 *
+	 * @param mimeType
+	 *            the {@link IMimeType} (never <i>null</i>)
+	 * @return this
+	 */
+	public AGStoredFile setContentTypeByMimeType(IMimeType mimeType) {
```
oliver-richers, line 220: why not call it just `setContentType`, otherwise its existence may be overlooked

oliver-richers (changes requested):

platform-core-module/src/main/java/com/softicar/platform/core/module/file/stored/AGStoredFile.java
alexander-schmidt: ok

oliver-richers (approved):

PR #699, opened 2023-02-06
EXP-2545 Implement IEmail.toMimeMessage()

oliver-richers:


platform-core-module/src/main/java/com/softicar/platform/core/module/email/buffer/BufferedEmailToMimeMessageConverter.java
```diff
@@ -0,0 +1,93 @@
...
+import jakarta.mail.BodyPart;
+import jakarta.mail.Message.RecipientType;
+import jakarta.mail.MessagingException;
+import jakarta.mail.Session;
+import jakarta.mail.internet.MimeBodyPart;
+import jakarta.mail.internet.MimeMessage;
+import jakarta.mail.internet.MimeMultipart;
+import jakarta.mail.util.ByteArrayDataSource;
+
+public class BufferedEmailToMimeMessageConverter {
```
alexander-schmidt, line 17: I assume this was tested _somehow_, e.g. via prototyping in a `main` method.
Considering out current, heavy time constraints: Could we at least turn the test code we already have into a unit test? That'd be better than nothing.

alexander-schmidt (changes requested):

platform-core-module/src/main/java/com/softicar/platform/core/module/email/buffer/BufferedEmailToMimeMessageConverter.java
oliver-richers: not really tested it only through UI, will write test

PR #720, opened 2023-02-13
PLAT-1318 Program executions: Add a 'Status' column

alexander-schmidt:
image
----
image
----
image

### DIFF
--- a/platform-core-module/src/main/resources/com/softicar/platform/core/module/program-execution-style.css
+++ b/platform-core-module/src/main/resources/com/softicar/platform/core/module/program-execution-style.css
@@ -1,6 +1,48 @@
+:root {
+	--PROGRAM_EXECUTION_COLOR_ERROR: #dc3545;
+	--PROGRAM_EXECUTION_COLOR_INFO: #407fbf;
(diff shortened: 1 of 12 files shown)


platform-core-module/src/main/resources/com/softicar/platform/core/module/program-execution-style.css
```diff
@@ -1,6 +1,48 @@
+:root {
+	--PROGRAM_EXECUTION_COLOR_ERROR: #dc3545;
+	--PROGRAM_EXECUTION_COLOR_INFO: #407fbf;
+	--PROGRAM_EXECUTION_COLOR_OKAY: #4bb91e;
```
oliver-richers, line 4: IMO should be using our default color for these states

oliver-richers (changes requested):

platform-core-module/src/main/resources/com/softicar/platform/core/module/program-execution-style.css
alexander-schmidt: ok

oliver-richers (approved):

PR #735, opened 2023-03-09
EXP-2731 No Workflow Notification by Default and Improved Error Output of BufferedEmailSender

oliver-richers:


daniel-klose (dismissed):

platform-db/src/main/java/com/softicar/platform/db/runtime/table/IDbTable.java
```diff
@@ -101,6 +102,19 @@
 	 */
 	R get(P primaryKey);
 
+	/**
+	 * Same as {@link #get} but wraps the result with
+	 * {@link Optional#ofNullable}.
+	 *
+	 * @param primaryKey
+	 *            the primary key of the row (may be null)
```
alexander-schmidt, line 110:
```suggestion
	 *            the primary key of the row (may be <i>null</i>)
```

alexander-schmidt (changes requested):

PR #736, opened 2023-03-09
PLAT-1331-2-original Enable StoredFileFileSystemContentStore

alexander-schmidt:
- Stored Files can now be saved to either SMB shares or local file system directories.
  - Renamed `Core`.`StoredFileServer` to `StoredFileRepository`. "Server" was just too confusing, now that we support `file:///` URLs. That renaming is reflected throughout both the UI and codebase.
  - An appropriate `IStoredFileContentStore` implementation is chosen according to the protocol part of the repository URL.
  - e.g. `StoredFileSmbContentStore` will be used for URLs like `smb://host/share`
  - e.g. `StoredFileFileSystemContentStore` will be used for URLs like `file:///dir/subdir`
- Established `StoredFileContentStores` as a central place to retrieve (SMB-based or file-system-based) `IStoredFileContentStore` instances.
  - This puts an end to seemingly-random calls to file store implementation constructors throughout the code base.
- `IStoredFileContentStore`: merged `isEnabled`, `isReady` and `isAvailable` to just one method, `isAvailable`.
  - All existing implementations of `isEnabled` were either trivial or nonsensically-arbitrary.
- `StoredFilesCleanupProgram`:
  - Now uses `StoredFileChunksToFileStoreMigrator` to migrate chunks to the currently preferred file store, instead of assuming an SMB based primary file store.
- `StoredFileContentInputStreamCreator` now checks whether a file store is available before trying to access it.


platform-core-module/src/main/java/com/softicar/platform/core/module/file/stored/content/store/StoredFileContentStoreFactory.java
```diff
@@ -0,0 +1,61 @@
...
+				Log.ferror("Cannot handle content store URLs that start with '%s'.", protocol);
+			}
+		} catch (MalformedURLException exception) {
+			Log.ferror("Invalid file repository URL: %s", StackTraceFormatting.getStackTraceAsString(exception));
+		}
+
+		return Optional.empty();
+	}
+
+	private static class SimpleUrl {
```
alexander-schmidt, line 33: `java.net.URL` throws for URLs that start with `smb`, so I built this.

platform-core-module/src/main/java/com/softicar/platform/core/module/file/stored/content/store/StoredFileContentStoreFactory.java
```diff
@@ -0,0 +1,61 @@
...
+		public SimpleUrl(String text) throws MalformedURLException {
+
+			var matcher = PATTERN.matcher(text);
+			matcher.find();
+			if (matcher.groupCount() != 2) {
+				throw new MalformedURLException("Illegal URL: '%s'".formatted(text));
+			}
+
+			this.protocol = matcher.group(1).toLowerCase();
+			this.path = Trim.trimPrefix(matcher.group(2), "//");
```
alexander-schmidt, line 48: Note that the following URLs are all valid, in principle:
```
file:/dir/subdir    - an absolute path
file://dir/subdir   - a relative path
file:///dir/subdir  - an absolute path
```

platform-dom/src/main/resources/com/softicar/platform/dom/dom-wiki-elements-style.css
```diff
@@ -4,12 +4,16 @@ h6.DomWikiHeadline {
...
+	margin-top: 0;
+}
+
 .DomWikiMonospace {
 	display: inline;
 	color: #880000;
 	background-color: #f0f0f0;
 	border-radius: var(--BORDER_RADIUS);
-	box-sahdow: 0 0 0 1px #cccccc inset;
+	box-shadow: 0 0 0 1px #cccccc inset;
```
alexander-schmidt, line 16: Fixed random typo / broken declaration

platform-dom/src/main/resources/com/softicar/platform/dom/dom-wiki-elements-style.css
```diff
@@ -4,12 +4,16 @@ h6.DomWikiHeadline {
 	border-bottom: 1px solid var(--LABEL_BORDER_COLOR);
 }
 
+.DomWikiList {
+	margin-top: 0;
```
alexander-schmidt, line 8: Before:
image

After:
image

platform-integration/src/main/java/com/softicar/platform/integration/database/structure/version/PlatformDatabaseStructureVersionResource.java
```diff
@@ -50,6 +50,6 @@ public interface PlatformDatabaseStructureVersionResource {
 	IResourceSupplier V31_STRUCTURE = FACTORY.create("v31-structure.json");
 	IResourceSupplier V32_MIGRATION = FACTORY.create("v32-migration.sql");
 	IResourceSupplier V32_STRUCTURE = FACTORY.create("v32-structure.json");
-	IResourceSupplier V34_MIGRATION = FACTORY.create("v34-migration.sql");
-	IResourceSupplier V34_STRUCTURE = FACTORY.create("v34-structure.json");
+	IResourceSupplier V38_MIGRATION = FACTORY.create("v38-migration.sql");
```
alexander-schmidt, line 53: v34 migration was empty, so it was promoted to v38

platform-core-module/src/main/java/com/softicar/platform/core/module/file/stored/content/store/StoredFileFileSystemContentStore.java
```diff
@@ -27,34 +28,30 @@ public class StoredFileFileSystemContentStore implements IStoredFileContentStore
...
-	}
-
-	@Override
-	public String getLocation() {
-
-		try {
-			return directory.getCanonicalPath();
-		} catch (IOException exception) {
-			throw new SofticarIOException(exception);
+		if (!absolutePath.startsWith("/")) {
```
daniel-klose, line 40: Hm, absoulte paths in windows always start with the drive, e.g. `C:\\`
Do we need to check OS here, or am I missing something?

platform-core-module/src/main/java/com/softicar/platform/core/module/file/stored/content/store/StoredFileContentStoreFactory.java
daniel-klose: But what about Windows paths?
Windows path always start with the drive `C:\\`

alexander-schmidt: I think this is irrelevant here. Who would create an `AGStoredFileRepository` with a Windows style directory URL, and why? I think we should deal with that when (if) problems occur.

platform-core-module/src/main/java/com/softicar/platform/core/module/file/stored/content/store/StoredFileFileSystemContentStore.java
alexander-schmidt: On Windows systems, we could use file URIs as exemplified here:
https://en.wikipedia.org/wiki/File_URI_scheme#Windows
I see no need to support a backslash based notation.

daniel-klose (approved):

PR #763, opened 2023-04-06
EXP-2827 Worfklow: Make finalized versions editable

daniel-klose:
Removed the nuance that "finalized" workflows cant be edited in any way.
Reasoning: Permissions to edit and create are already set to Administrators, and they should know what to do.


platform-workflow-module/src/main/java/com/softicar/platform/workflow/module/workflow/WorkflowPredicates.java
```diff
@@ -11,9 +11,6 @@ public interface WorkflowPredicates {
 		WorkflowI18n.DRAFT,
 		AGWorkflowVersion::isDraft);
 
-	// FIXME strange, not-draft implies finalized?
```
daniel-klose, line 14: This strangeness was finally killed, hooray

alexander-schmidt (approved):

PR #764, opened 2023-04-11
EXP-2801 Fix deferred initializers not being processed upon initial document creation

alexander-schmidt:
**Before:**
Peek 2023-04-11 18-09 before

---

**After:**
Peek 2023-04-11 18-08 after

### DIFF
--- a/platform-ajax/src/main/java/com/softicar/platform/ajax/document/creation/AjaxDocumentCreationService.java
+++ b/platform-ajax/src/main/java/com/softicar/platform/ajax/document/creation/AjaxDocumentCreationService.java
@@ -44,6 +44,8 @@ public class AjaxDocumentCreationService extends AbstractAjaxDocumentService {
 		document.getEngine().registerCssResourceLink(AjaxCssFiles.AJAX_STYLE.getResource());
 		document.getEngine().registerCssResourceLink(DomCssFiles.DOM_STYLE.getResource());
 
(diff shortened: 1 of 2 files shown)


platform-ajax/src/main/java/com/softicar/platform/ajax/document/creation/AjaxDocumentCreationService.java
```diff
@@ -44,6 +44,8 @@ private void buildDocument() {
 		document.getEngine().registerCssResourceLink(AjaxCssFiles.AJAX_STYLE.getResource());
 		document.getEngine().registerCssResourceLink(DomCssFiles.DOM_STYLE.getResource());
 
+		document.getDeferredInitializationController().handleAllAppended();
```
alexander-schmidt, line 47: Note that we still have a second call to `IDomDeferredInitializationController#handleAllAppended` which is in `DomEventHandlerNodeCaller#call`. That one is used whenever `AjaxDomEventService` handles an event.

I theory, those two calls could be unified to a single one in `AbstractAjaxDocumentService` but that would entail quite some refactoring which is not worth the effort.

oliver-richers (changes requested):
Can you test this, and append some videos?

alexander-schmidt:
Something went wrong with saving the attached GIFs. Should be visible now.

daniel-klose (approved):

PR #765, opened 2023-04-13
PLAT-1355 Reset program states before entering the QueuedProgramExecutionDaemon main loop

vdcstr:
This PR fixes situation no. 4 in the listing below.

```
Definitions:
- Execution := a record in AGProgramExecution
- State := a record in AGProgramState

There are 4 basic situations:

1: (Execution exists) && !(State references Execution)
=> already handled by ProgramExecutionsCleanupProgram

2: (Execution exists) && (State references Execution) && (Thread exists)
=> nothing to do, Thread's running normally

3: (Execution exists) && (State references Execution) && !(Thread exists) && (Thread not yet created)
=> nothing to do: that's expected before Thread startup

4: (Execution exists) && (State references Execution) && !(Thread exists) && (Thread is gone)
=> before this PR: ProgramExecutionsCleanupProgram would fail to clean up the Execution, and
   the program would be stuck in "orphaned" state until manual intervention
=> since this PR: ProgramExecutionsCleanupProgram will handle that because State is now reset
   for all programs upon QueuedProgramExecutionDaemon startup, which includes setting
   AGProgramState.currentExecution = NULL
```

Note that the setup operation in `QueuedProgramExecutionDaemon` is performed once per deamon creation. This happens everytime the Daemon Watchdog is (re-)started, which ofc. includes the initial servlet context startup.

### DIFF
--- a/platform-core-module/src/main/java/com/softicar/platform/core/module/daemon/IDaemon.java
+++ b/platform-core-module/src/main/java/com/softicar/platform/core/module/daemon/IDaemon.java
@@ -14,8 +14,8 @@ package com.softicar.platform.core.module.daemon;
 public interface IDaemon {
 
 	/**
-	 * Executes setup operations to perform before entering the daemon's main
-	 * loop.
+	 * Executed once before the daemon's main loop is entered, i.e. before the
+	 * very first call to {@link #runIteration()}.
 	 * <p>
 	 * Does nothing by default.
 	 */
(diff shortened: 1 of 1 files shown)


platform-core-module/src/main/java/com/softicar/platform/core/module/daemon/IDaemon.java
```diff
@@ -13,6 +13,17 @@
 @FunctionalInterface
 public interface IDaemon {
 
+	/**
+	 * Executes setup operations to perform before entering the daemon's main
+	 * loop.
```
oliver-richers, line 18: We should be more specific. Is this called only once, or before every `runIteration()`...same goes for the `cleanup()` method.

oliver-richers (dismissed):

platform-core-module/src/main/java/com/softicar/platform/core/module/daemon/IDaemon.java
alexander-schmidt: done

PR #766, opened 2023-04-13
PLAT-1355-2 Call ProgramExecutionsCleaner from a deamon

alexander-schmidt:
- Before this PR, `ProgramExecutionsCleaner` was called from a program which needs to be scheduled.
- That was inappropriate because `QueuedProgramExecutionDaemon` cannot properly function when `ProgramExecutionsCleaner` was not executed.
- One execution of `ProgramExecutionsCleaner` before entering the main loop of `QueuedProgramExecutionDaemon` should be sufficient.

Therefore:
- Integrated the program state reset into `ProgramExecutionsCleaner`.
- `ProgramExecutionsCleaner` is now called during the setup of `QueuedProgramExecutionDaemon`, and only there.


platform-core-module/src/main/java/com/softicar/platform/core/module/program/execution/cleanup/ProgramExecutionsCleanupProgram.java
```diff
@@ -27,7 +27,6 @@ public class ProgramExecutionsCleanupProgram implements IProgram {
 	@Override
 	public void executeProgram() {
 
-		new ProgramExecutionsCleaner().cleanupOrphanedRecords();
```
oliver-richers, line 30: - second paragraph in Javadoc must be removed
- class should be renamed to `ProgramExecutionsDeleterProgram`, maybe rename both classes:
  - `OldProgramExecutionsDeleterProgram`
  - `OldProgramExecutionsDeleter`

platform-core-module/src/test/java/com/softicar/platform/core/module/program/execution/ProgramExecutionsCleanerTest.java
```diff
@@ -22,21 +21,24 @@ public ProgramExecutionsCleanerTest() {
...
-		insertProgramState(program, activeExecution);
+		AGProgramExecution executionWithState = insertProgramExecution();
+		insertProgramState(program, executionWithState);
+		assertFalse(executionWithState.isFailed());
+		assertNull(executionWithState.getTerminatedAt());
 
-		AGProgramExecution orphanedExecution = insertProgramExecution();
-		assertFalse(orphanedExecution.isFailed());
-		assertNull(orphanedExecution.getTerminatedAt());
+		AGProgramExecution executionWithoutState = insertProgramExecution();
```
oliver-richers, line 29: why you renamed the variable but not the `cleanupOrphanedExecutions` method?

oliver-richers (changes requested):

platform-core-module/src/main/java/com/softicar/platform/core/module/program/execution/cleanup/ProgramExecutionsCleanupProgram.java
alexander-schmidt: ok

oliver-richers (approved):

PR #780, opened 2023-04-21
PLAT-1354 Badges for Page Navigation Links

oliver-richers:
https://user-images.githubusercontent.com/83839745/233652357-611edecb-fe63-4cda-8ac8-285d55d12912.mp4


platform-core-module/src/main/resources/com/softicar/platform/core/module/page-navigation-style.css
```diff
@@ -5,6 +5,31 @@
...
+	border-radius: 100%;
+	margin-left: 0.2em; 
+	min-width: 1em;
+	padding: 0.2em;
+	text-align: center;
+}
+
+.PageNavigationBadge.error {
+	background-color: var(--ERROR_BORDER_COLOR);
+	color: #fff;
```
daniel-klose, line 20: if the color is always the same, why not move it to `.PageNavigationBadge`?

platform-emf/src/main/java/com/softicar/platform/emf/page/EmfPageBadge.java
```diff
@@ -0,0 +1,141 @@
...
+	/**
+	 * Constructs this badge with the given {@link EmfDiagnosticLevel} and count
+	 * supplier.
+	 *
+	 * @param level
+	 *            the {@link EmfDiagnosticLevel} (never <i>null</i>)
+	 * @param countSupplier
+	 *            the {@link Supplier} for the badge counter (never <i>null</i>)
+	 */
+	public EmfPageBadge(EmfDiagnosticLevel level, Supplier<Integer> countSupplier) {
```
daniel-klose, line 31: Hm, maybe we need a title for the badges that pops up when you hover over them?
Currently badges are just arbitrary numbers with different colored backgrounds that dont really tell anything

(Granted, current implementation is self explanatory, but this could be used in more pages in the future)

platform-core-module/src/main/java/com/softicar/platform/core/module/maintenance/MaintenanceWindowPage.java
```diff
@@ -37,4 +43,21 @@ public EmfPagePath getPagePath(EmfPagePath modulePath) {
...
+			.where(AGMaintenanceWindow.STATE.isEqual(AGMaintenanceStateEnum.IN_PROGRESS.getRecord()))
+			.count();
+		Supplier<Integer> pendingCountSupplier = () -> AGMaintenanceWindow.TABLE//
+			.createSelect()
+			.where(AGMaintenanceWindow.STATE.isEqual(AGMaintenanceStateEnum.PENDING.getRecord()))
+			.count();
+		return List
+			.of(//
+				new EmfPageBadge(EmfDiagnosticLevel.WARNING, inProgressCountSupplier).setRefreshClasses(AGMaintenanceWindow.class),
+				new EmfPageBadge(EmfDiagnosticLevel.INFO, pendingCountSupplier).setRefreshClasses(AGMaintenanceWindow.class));
```
daniel-klose, line 61: How does this look?
one yellow badge with a number and a blue one with a number?
If so, see comment further down about labeling

daniel-klose (changes requested):

platform-core-module/src/main/java/com/softicar/platform/core/module/page/navigation/entry/link/PageNavigationBadge.java
```diff
@@ -0,0 +1,51 @@
...
+import com.softicar.platform.core.module.CoreCssClasses;
+import com.softicar.platform.dom.DomCssPseudoClasses;
+import com.softicar.platform.dom.elements.DomDiv;
+import com.softicar.platform.dom.refresh.bus.IDomRefreshBusEvent;
+import com.softicar.platform.dom.refresh.bus.IDomRefreshBusListener;
+import com.softicar.platform.dom.style.ICssClass;
+import com.softicar.platform.emf.page.EmfPageBadge;
+import com.softicar.platform.emf.validation.result.EmfDiagnosticLevel;
+
+public class PageNavigationBadge extends DomDiv implements IDomRefreshBusListener {
```
alexander-schmidt, line 12: We have `PageNavigationBadge` and `EmfPageBadge`. The former is a DOM element while the latter is not. Calling both of those a `Badge` appears odd to me.
In the past and in similar situations, we appended a `Definition` suffix, e.g. in `DomImageViewerTagDefinition` or `IDaemonDefinition`.

platform-core-module/src/main/resources/com/softicar/platform/core/module/page-navigation-style.css
```diff
@@ -5,6 +5,31 @@
 	--PAGE_NAVIGATION_ENTRY_HEIGHT: 32px;
 }
 
+/* -------------------- PageNavigationBadge -------------------- */
+
+.PageNavigationBadge {
```
alexander-schmidt, line 10: Does this look alright for 3- or even 4-digit numbers?

platform-emf/src/main/java/com/softicar/platform/emf/page/EmfPageBadge.java
alexander-schmidt, line 31: We currently do not tell the user what the displayed number refers to, in any way. When the page is opened, this might not be obvious.
Instead of letting users guess, I think we should require an `IDisplayString`-based explanatory text as an additional parameter.

platform-core-module/src/main/resources/com/softicar/platform/core/module/page-navigation-style.css
oliver-richers: ok

platform-core-module/src/main/java/com/softicar/platform/core/module/maintenance/MaintenanceWindowPage.java
oliver-richers: yes, see the demo video

platform-emf/src/main/java/com/softicar/platform/emf/page/EmfPageBadge.java
oliver-richers: did so
image

alexander-schmidt (approved):

PR #782, opened 2023-04-24
PLAT-1363 Set SMTP properties as String

alexander-schmidt:
- Before this PR, we set SMTP timeout properties as `long` values.
- A literal interpretation of the documentation at (https://javaee.github.io/javamail/docs/api/com/sun/mail/smtp/package-summary.html) implies that properties need to be set as String:
  > The SMTP protocol provider supports the following properties, which may be set in the JavaMail Session object. The properties are always set as strings; the Type column describes how the string is interpreted.

Unrelated:
- Slightly improved log output.


oliver-richers (approved):
After checking the implementation, we discovered that Integer and String are fine, but not Long, which we previously used. This effectively caused the interpretation as infinite timeout.
image

PR #794, opened 2023-05-15
PLAT-1373 ExceptionsCollector: Throw on collecting InterruptedException

alexander-schmidt:


platform-common/src/main/java/com/softicar/platform/common/core/exception/ExceptionsCollector.java
```diff
@@ -30,6 +31,10 @@ public ExceptionsCollector setPreamble(String preamble, Object...arguments) {
 
 	/**
 	 * Adds the given {@link Exception} to this {@link ExceptionsCollector}.
+	 * <p>
+	 * If the given {@link Exception} is an {@link InterruptedException} or
+	 * caused by one, the added Exception will be wrapped in a
```
oliver-richers, line 36:
```suggestion
	 * caused by one, the added {@link Exception} will be wrapped in a
```
or just
```suggestion
	 * caused by one, it will be wrapped in a
```

oliver-richers (approved):

PR #809, opened 2023-05-24
PLAT-1375-2 I18n translated to Bosnian, Croatian and Serbian

semsudin:


oliver-richers (approved):
I am not an expert in these languages, but seems ok.

PR #824, opened 2023-05-31
PLAT-1365 Filter Entity Column by Text in EMF Data Table

oliver-richers:
https://github.com/softicar/platform/assets/83839745/12325399-4634-421b-985f-987c5bc5307e

### DIFF
--- a/platform-emf/src/main/java/com/softicar/platform/emf/data/table/filter/entity/EmfDataTableEntityFilterNode.java
+++ b/platform-emf/src/main/java/com/softicar/platform/emf/data/table/filter/entity/EmfDataTableEntityFilterNode.java
@@ -40,12 +40,6 @@ public class EmfDataTableEntityFilterNode<R, T extends IEntity> extends Abstract
 	}
 
 	@Override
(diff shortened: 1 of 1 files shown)


platform-emf/src/main/java/com/softicar/platform/emf/data/table/filter/entity/EmfDataTableEntityFilterNode.java
```diff
@@ -1,79 +1,47 @@
...
-			} else if (filterType.isNotEmpty()) {
-				return new EmfDataTableNotEmptyValueFilter<>(column.getDataColumn(), resetter);
-			}
-			return new EmfDataTableNopFilter<>(resetter);
-		}
+		return new EmfDataTableEntityFilterFactory<>(this).createFilter();
 	}
 
 	@Override
 	public void refresh() {
```
daniel-klose, line 42: Since we only call super.refresh, cant we remove this override altogether?

daniel-klose (approved):

PR #843, opened 2023-06-27
PLAT-1382-101 Login Page set max width for logo

semsudin:


emirPuskaForspace (approved):

platform-core-module/src/main/java/com/softicar/platform/core/module/page/service/login/PageServiceLoginDiv.java
```diff
@@ -85,7 +87,7 @@ public LoginForm() {
 
 		private void appendLogo(AGStoredFile logoFile) {
 
-			appendChild(new DomImage(new StoredFileResource(logoFile)));
+			appendChild(new DomImage(new StoredFileResource(logoFile)).setStyle(CssStyle.MAX_WIDTH, new CssPixel(250)));
```
oliver-richers, line 90: This should not be hard-coded in Java, but in the CSS Style.

oliver-richers (changes requested):

PR #846, opened 2023-06-27
PLAT-1386 SystemUser for AutoTransitions in MessageHistory

daniel-klose:


platform-workflow-module/src/main/java/com/softicar/platform/workflow/module/workflow/item/message/WorkflowItemMessageRow.java
```diff
@@ -48,6 +52,9 @@ public DayTime getCreatedAt() {
 
 	public AGUser getCreatedBy() {
 
+		if (isAutoTransitionMessage) {
```
daniel-klose, line 55: This is pretty hacky, but it was the best solution I could come up with after hours of trying to understand the whole message mechanisms and trying different things where nothing worked....

platform-workflow-module/src/main/java/com/softicar/platform/workflow/module/workflow/item/message/WorkflowItemMessageRowsLoader.java
```diff
@@ -128,7 +128,7 @@ private WorkflowItemMessageRow createAutoTransitionSourceRow(AGWorkflowAutoTrans
 
 		return createTransitionRow(//
 			transitionExecution.getWorkflowTransition().getSourceNode(),
-			transitionExecution.getTransaction());
+			transitionExecution.getTransaction()).setIsAutoTransitionMessage(true);
```
oliver-richers, line 131: - This won't work as expected, it will make the last auto transition seem to be done by the user and the first transition to be done by the system user, which is both wrong. You can also see this in the second screenshot.
- We need to talk.

oliver-richers (changes requested):

PR #850, opened 2023-06-29
PLAT-146 Add password reset in login form

daniel-klose:


oliver-richers (changes requested):
As discussed, needs to go over email and needs to be limited per day.

platform-core-module/src/main/java/com/softicar/platform/core/module/user/password/UserPasswordGenerator.java
```diff
@@ -54,10 +56,17 @@ public void resetUserPassword(AGUser user) {
 		try (DbTransaction transaction = new DbTransaction()) {
 			String password = new UserPasswordGenerator().generatePassword();
 			new UserPasswordUpdater(user, password).update();
-			AGUser.sendPaswordResetNotification(user, password);
+			if (AGCoreModuleInstance.getInstance().getEmailServer() != null) {
+				AGUser.sendPaswordResetNotification(user, password);
+			}
```
oliver-richers, line 61: shouldn't we assume that the email server is configures? otherwise, the whole password reset mechanism would not work

platform-core-module/src/main/java/com/softicar/platform/core/module/user/password/reset/PasswordResetDiv.java
```diff
@@ -0,0 +1,214 @@
...
+import com.softicar.platform.dom.elements.message.DomMessageDiv;
+import com.softicar.platform.dom.elements.message.style.DomMessageType;
+import com.softicar.platform.dom.event.DomEventType;
+import com.softicar.platform.dom.event.IDomClickEventHandler;
+import com.softicar.platform.dom.event.IDomEvent;
+import com.softicar.platform.dom.input.AbstractDomInput;
+import com.softicar.platform.dom.styles.CssDisplay;
+import java.util.Optional;
+
+public class PasswordResetDiv extends DomDiv {
```
oliver-richers, line 36: I really wonder it we need this whole DIV at all. Wouldn't it suffice to simply generate a new password and send it per mail to the user? That's the way, how the admin does the password reset.

oliver-richers (changes requested):

platform-core-module/src/main/java/com/softicar/platform/core/module/user/password/reset/PasswordResetDiv.java
daniel-klose: I do strongly belive so, yes.
Every (good) password reset mechanism I know allows users to input their new password themselves, exactly like this form.

platform-core-module/src/main/java/com/softicar/platform/core/module/CoreI18n.java
```diff
@@ -1034,12 +1044,17 @@ public interface CoreI18n extends EmfI18n {
 		.bs("Sat")
 		.sr("Sat")
 		.hr("Sat");
+	I18n0 IF_AN_EMAIL_SERVER_IS_CONFIGURED_THE_USER_WILL_BE_NOTIFIED_VIA_EMAIL =
```
oliver-richers, line 1047: can you scan this class for dead entries?

platform-core-module/src/main/java/com/softicar/platform/core/module/user/password/UserPasswordGenerator.java
```diff
@@ -42,7 +44,7 @@ public String generatePassword() {
 		throw new SofticarUserException(CoreI18n.FAILED_TO_GENERATE_PASSWORD);
 	}
 
-	public void resetUserPassword(AGUser user) {
+	public void resetUserPassword(AGUser user, boolean showNewPassword) {
```
oliver-richers, line 47: where is the call for this method with parameter `false`?

platform-core-module/src/main/java/com/softicar/platform/core/module/user/password/UserPasswordGenerator.java
```diff
@@ -52,12 +54,18 @@ public void resetUserPassword(AGUser user) {
...
+				firstMessagePart = CoreI18n.THE_PASSWORD_FOR_USER_ARG1_IS_NOW_ARG2.toDisplay(user.getLoginName(), password);
+			} else {
+				firstMessagePart = CoreI18n.THE_PASSWORD_FOR_USER_ARG1_HAS_BEEN_RESET.toDisplay(user.getLoginName());
+			}
 			new DomModalAlertDialog(
-				CoreI18n.THE_PASSWORD_FOR_USER_ARG1_IS_NOW_ARG2
-					.toDisplay(user.getLoginName(), password)
+				firstMessagePart//
 					.concat("\n\n")
 					.concat(CoreI18n.IF_AN_EMAIL_SERVER_IS_CONFIGURED_THE_USER_WILL_RECEIVE_THIS_PASSWORD_VIA_EMAIL)).open();
```
oliver-richers, line 70: we don't need the if-part of this sentence if we throw an exception in line 57....but actually, I don't know what this method has to do with the whole "forgot password" mechanism??? there is only one caller for the `resetUserPassword` method

oliver-richers (changes requested):

platform-core-module/src/main/java/com/softicar/platform/core/module/user/password/UserPasswordGenerator.java
daniel-klose: Its the user password reset action, currently admin get shown the new password.
To preserve this, I added this flag

platform-core-module/src/main/java/com/softicar/platform/core/module/user/password/UserPasswordGenerator.java
daniel-klose: oh right, I used this in an old version and forgot to revert the changes
Reverted everything here

oliver-richers (approved):

PR #857, opened 2023-08-07
PLAT-1418: Added LengthBits validation to EmfStringAttribute

daniel-klose:


oliver-richers (changes requested):
- I don't think, we can trust the documentation on this, because it does not make sense. I will investigate this.
- We should also have a test that really tests the difference between bytes and characters.

oliver-richers (changes requested):
- According to the MariaDB documentation, the maximum length is to be considered in bytes, not characters:
grafik

- So, IMO, we should take the text, convert it to the actual character set of the column and then check the text length in bytes.
- We could also simplify this and always convert to utf8 as a pessimistic simplification.

oliver-richers (changes requested):

oliver-richers (approved):

